PR #11, opened 2024-08-22
feat: deal with qpdf errors on a page by page basis

PeterStaar-IBM:


dolfim-ibm (approved):
The situation is:

If anything happen `pages` will **not** be there, so it can be checked.

PR #13, opened 2024-08-23
ci: build in Release mode and improve compilation with multi-threading

dolfim-ibm:
- Set `CMAKE_BUILD_TYPE=Release`
- Avoid building twice the code from setup-poetry
- Control the number of threads for building via ENV


PeterStaar-IBM (approved):
Looks got to me!

PR #17, opened 2024-09-09
feat: build linux arm64 architecture

dolfim-ibm:


dolfim-ibm:
It looks like `cibuildwheel` is using the qemu emulation for building linux aarch64. This makes it very slow to build, i.e. it takes > 45min.

PR #28, opened 2024-09-27
fix: sdist and wheels content

### LINKED ISSUE #24: missing sdist in pypi index
Would it be possible to start publising sdists for `docling-parse` along with the wheels on the pypi index? We are trying to package Docling for [InstructLab](https://github.com/instructlab) and need the sources to build it.

### LINKED ISSUE #25: installing `pybind11` as a build time python package dependency
When we try to build `docling-parse` from source, it tries to clone the `pybind11` repository and then set the includes path. However, we are running this build in a network isolated environment which causes the clone step to fail.

An alternate to this would be to install `pybind11` as a build time python package dependency and use its API in `build.py` to set it up:
- https://pybind11.readthedocs.io/en/stable/compiling.html#modules-with-setuptools
- https://github.com/octavifs/poetry-pybind11-integration

### LINKED ISSUE #27: Windows wheel is not working
The published windows wheel is buggy
1. it missing the actual compiled library. This is very likely excluded from the `include={}` list.
2. the wheel (when found) is linking dynamically to qpdf

dolfim-ibm:
Resolves #24 
Resolves #25 
Resolved #27

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -116,7 +116,7 @@ jobs:
             echo "Inspecting $file"
             poetry run python -m zipfile --list "$file"
           done
-          mkdir ./dist
+          mkdir -p ./dist
           cp wheelhouse/*.whl ./dist/
 
       # there is an error with the tagging of wheels for macosx-arm64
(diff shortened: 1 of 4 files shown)


PeterStaar-IBM (approved):
LGTM!

PR #29, opened 2024-09-28
feat: add a v2 parser to improve performance

staar:


cau-git:
Some observations from testing:
- A few files from our test set cause fatal aborts (`SIGABRT`), which can't be caught on python. Details follow.
- Parsing PDFs page-by-page is very fast, but it takes successively longer for every additional page in a document, and the output for page N contains the content of pages 0 to N-1 as well. This also grows memory usage.
- The parser_v2 should accept the loglevel straight at construction time (on `__init__`), so it doesn't spill logs before user code can call `parser.set_loglevel` to silence it.

dolfim-ibm (approved):

PR #31, opened 2024-10-01
fix: python version in wheels

dolfim-ibm:
- [x] Fix the duplicate publish of sdist
- [x] Fix wheels containing wrong python modules
- Python 3.9 CI is disabled, since there is no pairing docling version for it
- More tests
    - Validate that poetry using the right python version
    - Validate that the .so is contained in the wheels
    - Validate that the wheels contain the expected python version
    - Try using docling-parse with the latest docling


PeterStaar-IBM (approved):
Let's go!

PR #41, opened 2024-10-17
fix: fatal errors on pdfs

PeterStaar-IBM:


cau-git (approved):

cau-git:
Verified through docling with previously broken test samples. No crashes, and we get at least partial output now. Good to merge.

PR #42, opened 2024-10-17
Fix cmake segfault by renaming ext_name_cxxopts

tiran:
Our builds are failing with a segfault with cmake 3.26.5 on RHEL. The final phase of CMake is trapped in an infinitive recursion and eventually exhausts all stack trace. The segfault only occurs with `USE_SYSTEM_DEPS=ON`.

Eventually we found the culprit in `extlib_cxxopts.cmake`. The line `add_dependencies(${ext_name_cxxopts} cxxopts)` is causing the segfault. `ext_name_cxxopts` is `cxxopts`. I'm not entirely sure why it is failing. I think there is a variable name clash. Setting `ext_name_cxxopts` to a different value resolves the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ include(cmake/extlib_jpeg.cmake)
 include(cmake/extlib_qpdf_v11.cmake)
 
 # aggregate the targets created by the dependencies
-set(DEPENDENCIES qpdf jpeg utf8 json loguru cxxopts)
+set(DEPENDENCIES qpdf jpeg utf8 json loguru ext_cxxopts)
 
 # ************************
 # *** libraries        ***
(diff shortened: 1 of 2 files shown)


PeterStaar-IBM (approved):
Looks good to me!

tiran:
We made a mistake during our testing last night. This change does not fix the infinitive recursion segfault in cmake. [`extlib_loguru.cmake`](https://github.com/DS4SD/docling-parse/blob/main/cmake/extlib_loguru.cmake) is also causing trouble. I don't have a good understanding of cmake. Whatever you are doing in https://github.com/DS4SD/deepsearch-glm/tree/main/cmake works and whatever docling-parse is doing differently fails for us.

PeterStaar-IBM:
@tiran ok, let me take a look.

PeterStaar-IBM:
@tiran I did not want to pollute the branch of your fork. If we want to do something similar as glm, this is done in this new PR (https://github.com/DS4SD/docling-parse/pull/44)

PeterStaar-IBM:
@tiran shall we close this? I think it is now all resolved with #44, right?

tiran:
Yes, this PR is no longer needed with #44.

PR #44, opened 2024-10-18
fix: cmake-cxxopts by using similar approach as glm

PeterStaar-IBM:


tiran:
Thanks! Could you also copy the loguru cmake code? That seems to cause trouble, too.

PeterStaar-IBM:
@tiran yes, of couse! Can you confirm the cxxopts now works? If so, we might want to close your PR (#42 ) and get this one through asap.

PeterStaar-IBM:
@tiran loguru done: now lets test ;)

dolfim-ibm (approved):

PR #45, opened 2024-10-18
Fix `build_rhel.sh` to create isolated build

tiran:
- install DNF plugins for copr and config manager
- install missing `cxxopts-devel`
- install `loguru-devel` from EPEL 9. The update has landed two days ago.
- move `USE_SYSTEM_DEPS` into container file to address a backslash escape problem
- build the wheel in an isolated network namespace with `unshare -rn`. Since the build step is longer able to install its build dependencies, we have to pre-install the packages and use `--no-build-isolation`.

Tests will now fail with a segfault in `cmake` until #44 has landed.

### DIFF
--- a/.github/scripts/build_rhel.sh
+++ b/.github/scripts/build_rhel.sh
@@ -6,8 +6,6 @@ set -x  # display command on output
 # Build the Python package with Poetry
 poetry build -f sdist
 
-USE_SYSTEM_DEPS="ON"
-
 docker build --progress=plain \
              --build-arg USE_SYSTEM_DEPS="$USE_SYSTEM_DEPS" \
              -f - . <<EOF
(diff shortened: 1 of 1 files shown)


dolfim-ibm (approved):
Great!

tiran:
rebased

tiran:
Docker is blocking the "unshare" syscall. It's not a problem with podman. I'll get back to this next week.

dolfim-ibm:
@tiran [here](https://github.com/tiran/docling-parse/pull/1) I tried switching to Podman, but we still have some errors

```
sd-bus call: Transport endpoint is not connected
```

could this be caused by some permissions limitation of GH runner?

dolfim-ibm:
Completed in #63. Thanks a lot @tiran

PR #51, opened 2024-10-30
chore: update poetry in CI

dolfim-ibm:
This change should fix the list of python versions reported by PyPI.


PeterStaar-IBM (approved):
LGTM

PR #52, opened 2024-10-30
fix: improve qpdf optimization options

dolfim-ibm:
From the previous CI builds we noticed the following message.

```
    CMake Warning at CMakeLists.txt:405 (message):
        CMAKE_BUILD_TYPE is not set; using default settings
```

This was not adding any optimization flag to the QPDF compilation.

The additional parameter we are introducing is the same as the [qpdf packaging](https://github.com/qpdf/qpdf/blob/main/build-scripts/build-linux).


PeterStaar-IBM (approved):
lgtm!

PR #58, opened 2024-11-17
feat: add the export of annotations and ToC

PeterStaar-IBM:


PeterStaar-IBM:
We can now extract the fillable fields,

Screenshot 2024-11-18 at 06 00 42

ericfeunekes:
Will this PR include docs for usage?

PeterStaar-IBM:
@ericfeunekes No, just for regression tests.

src/v2/pdf_decoders/document.h
```diff
@@ -21,6 +21,9 @@ namespace pdflib
     nlohmann::json get();
 
     int get_number_of_pages() { return number_of_pages; }
+
+    nlohmann::json get_annotations() { return json_annots; }
+    nlohmann::json get_table_of_contents() { return json_annots["table_of_contents"]; }
```
maxmnemonic, line 26: "get_table_of_contents" is a convenience method here, as "get_annotations" will provide the same data right?
In this case should here be some sort of explicit check for "table_of_contents" key, because it's not guaranteed that we will have it?

dolfim-ibm (approved):
👍

PR #71, opened 2024-11-27
ci: add upload to build artifacts and push once together to pypi

### LINKED ISSUE #67: Improve CI flow
We are currently pushing to pypi from multiple pipelines, each taking care of its own branch (mac, win, linux, py versions, etc).

This creates some inconsistencies on pypi, because it could take >1h before all wheels are available. For example, we noticed multiple times that poetry would detect only part of the new version and we have to force a cache clear to see all of them.

The clean solution is to first push the artifacts as attachments in the GH release, then, as a single workflow, upload them together to pypi.

### LINKED ISSUE #68: Reduce number of mac builds
We are now building too many mac versions. It should be possible to have 1x x86_64 and 1x arm64 for each python version. We don't need specific macos versions.

dolfim-ibm:
Resolves #67.
Resolves #68.

In this PR we apply the following changes
1. Packages are first saved in the CI build artifacts, and then pushed to pypi all together.
    - these wheels will be saved also in PRs, so we can download them and try locally
3. We use the new pypi trusted publishers option for pushing to pypi
4. Packages are signed and published also as the GH release attachments
5. Mac releases now support 10.13+ with a single wheel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,9 @@ jobs:
   build-wheels:
     if: ${{ github.event_name == 'push' || (github.event.pull_request.head.repo.full_name != 'DS4SD/docling-parse' && github.event.pull_request.head.repo.full_name != 'ds4sd/docling-parse') }}
     uses: ./.github/workflows/wheels.yml
+    permissions:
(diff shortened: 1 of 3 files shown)


PeterStaar-IBM (approved):
Nice!!

PR #82, opened 2025-01-11
feat: Update for complex fonts, rendering, and experimental high-level API

PeterStaar-IBM:


cau-git:
I checked this with most of the known examples, and it appears to work much better!

However, there are other cases left where the output is still bad, I collected these below:

1. [EL00820224CPARE.PDF](https://github.com/user-attachments/files/18396294/EL00820224CPARE.PDF) - some text decodes with weird chars, e.g. `Leilªo  Pœblico` instead of `Leilão Público` (expected), or `Alienaçªo FiduciÆria` instead of `Alienação Fiduciária` (expected).
2. [glyph_text.pdf](https://github.com/user-attachments/files/18396343/glyph_text.pdf) - only glyph symbols are printed, in this form: `/G42/G79/G82/G69/G68/...`
3. [050216e467d363616161fb79e2bfaa3e.pdf](https://github.com/user-attachments/files/18396382/050216e467d363616161fb79e2bfaa3e.pdf) - many parts are glyphs, in a different form: `*OREDOLVLHUXQJGLYPH<3>LQGLYPH<3>GHUGLYPH<3>6SHLVHNDPPHU
%DQGGLYPH<3>GLYPH<20>GLYPH<29>`

cau-git (dismissed):

cau-git:
We discovered one regression: Some cells come out with zero-height now.

PR #86, opened 2025-01-17
fix: move autoflake to dev dependencies

hadim:


hadim:
Fix https://github.com/DS4SD/docling-parse/issues/84

dolfim-ibm (dismissed):

dolfim-ibm:
@hadim note that when touching the pyproject.toml file, you also have to run something like `poetry lock --no-update` for updating the lock file.

hadim:
@dolfim-ibm I also removed the entrypoint (which is missing from the source code) as said at https://github.com/DS4SD/docling-parse/issues/85

dolfim-ibm (approved):

dolfim-ibm:
> Maybe you should consider relying on the conda-forge deps for jpeg, qpdf and others instead in order to speedup the CI duration?

The actual long time is the cross-compilation for linux-arm64. For building the distribution wheels at the moment we require the static libraries of the dependencies, which I think are not available in the coda packages.

PR #87, opened 2025-01-23
fix: added more updates to better font-parsing

PeterStaar-IBM:


tests/test_parse_v1.py
```diff
@@ -87,7 +87,7 @@ def test_reference_documents_from_filenames_with_keys():
         rname = os.path.basename(pdf_doc)
         fname = os.path.join(GROUNDTRUTH_FOLDER, rname + ".v1.json")
 
-        if GENERATE:
+        if GENERATE or (not os.path.exists(fname)):
```
dolfim-ibm, line 90: we should also avoid that files are generated by CI

```suggestion
        if not os.getenv("CI") and (GENERATE or (not os.path.exists(fname))):
```

dolfim-ibm (approved):
tested in Docling and it works fine

dolfim-ibm (approved):

PR #103, opened 2025-02-14
ci: enable ppc64le wheel

mgiessing:
This PR enables building ppc64le manylinux wheels on GitHub Actions. It does not introduce any content changes.


cau-git:
IMG_8961

The ppc64le wheels seems to have succeeded in GitHub CI.

cau-git:
@mgiessing  yes, If we have now two emulated builds it would definitely make sense to split and parallelize. And we can adjust the naming of the build job for these.

Issue #99, opened 2025-02-06
Wrong text output


PeterStaar-IBM:
[Stafford.County.-.VA.Zoning.Ordinance.pdf](https://github.com/user-attachments/files/18687439/Stafford.County.-.VA.Zoning.Ordinance.pdf)

dhdaines:
I've bisected this on the document that was giving me the problem and it seems to be due to a change that happened between 3.1.2 and 3.2.0.  My document and the one above seem to have in common that their text objects consist of individual characters positioned absolutely, like this (you can use [`playa --content-streams`](https://github.com/dhdaines/playa) to see this easily 😉)

```
BT
/F4 10.6599998 Tf
1 0 0 -1 34.015625 29 Tm
<0016> Tj
5.9285736 0 Td <0012> Tj
2.9616852 0 Td <0014> Tj
5.9285736 0 Td <001B> Tj
5.9285736 0 Td <0012> Tj
2.9616852 0 Td <0015> Tj
5.9285736 0 Td <0017> Tj
5.9285736 0 Td <000F> Tj
2.9616852 0 Td <0003> Tj
2.9616852 0 Td <0014> Tj
5.9285736 0 Td <0015> Tj
5.9285736 0 Td <001D> Tj
2.9616852 0 Td <0015> Tj
5.9285736 0 Td <0019> Tj
5.9285736 0 Td <0003> Tj
2.9616852 0 Td <0033> Tj
7.1101227 0 Td <0030> Tj
ET
```

Probably there is some bug in the way docling-parse is handling the text matrix that is causing the characters to end up in the wrong place when they are positioned on the page.  They are nonetheless in the correct reading order in the content stream.

dhdaines:
Without actually building docling-parse here, if I had to guess, I'd bet that #91 is somehow responsible (but GitHub won't let me see the diff for some reason...)

dhdaines:
Confirmed that it is #91, when comparing `parse_v2.exe` output between 760b932b6770d928380ee83cdd9d14b901f695f8..v3.2.0 I get this sort of difference:
```
@@ -35424,7 +35512,139 @@
               648.32,
               37.074,
               648.32,
-              "Consult the following schematics for information about the various ports used by TrueSight Infrastructure Management. The port numbers",
[...]

PR #105, opened 2025-03-05
fix: handle case of overlapping cells in contract_cells_into_lines_v2

dhdaines:
Fixes: #99

As noted there, the (nicely refactored!) implementation no longer correctly handles the case where text cells overlap.

The fact that they overlap in the first place is most likely a separate and more complex bug, but this fixes the symptoms, at least.

### DIFF
--- a/src/v2/pdf_resources/page_cell.h
+++ b/src/v2/pdf_resources/page_cell.h
@@ -27,6 +27,10 @@ namespace pdflib
     
     bool is_adjacent_to(pdf_resource<PAGE_CELL>& other, double delta);
 
+    bool intersects(pdf_resource<PAGE_CELL>& other);
+
+    bool contains(double x, double y);
+
     bool has_same_reading_orientation(pdf_resource<PAGE_CELL>& other);
     
     bool merge_with(pdf_resource<PAGE_CELL>& other, double delta);
@@ -259,13 +263,64 @@ namespace pdflib
     
     return (num_chars>0? len/num_chars : 0.0);
   }
-  
+
+  bool pdf_resource<PAGE_CELL>::intersects(pdf_resource<PAGE_CELL>& other)
+  {
+    // Use point-in-polygon (via even-odd rule) to determine if
+    // bounding quadrilaterals intersect.
+    return (contains(other.r_x0, other.r_y0)
+            or contains(other.r_x1, other.r_y1)
+            or contains(other.r_x2, other.r_y2)
+            or contains(other.r_x3, other.r_y3)
+            or other.contains(r_x0, r_y0)
+            or other.contains(r_x1, r_y1)
+            or other.contains(r_x2, r_y2)
+            or other.contains(r_x3, r_y3));
+  }
+
+  inline bool inside_plane(double x, double y, double xi, double yi, double xj, double yj)
+  {
+    return ((yi > y) != (yj > y) and (x < (xj - xi) * (y - yi) / (yj - yi) + xi));
+  }
+
+  bool pdf_resource<PAGE_CELL>::contains(double x, double y)
+  {
+    // point-in-polygon via even-odd rule
+    bool inside = false;
+    if (inside_plane(x, y, r_x3, r_y3, r_x0, r_y0))
+      inside = not inside;
+    if (inside_plane(x, y, r_x0, r_y0, r_x1, r_y1))
+      inside = not inside;
+    if (inside_plane(x, y, r_x1, r_y1, r_x2, r_y2))
+      inside = not inside;
+    if (inside_plane(x, y, r_x2, r_y2, r_x3, r_y3))
+      inside = not inside;
+    return inside;
+  }
+
+
   bool pdf_resource<PAGE_CELL>::is_adjacent_to(pdf_resource<PAGE_CELL>& other, double eps)
   {
-    double d0 = std::sqrt((r_x1-other.r_x0)*(r_x1-other.r_x0) + (r_y1-other.r_y0)*(r_y1-other.r_y0));
-    double d1 = std::sqrt((r_x2-other.r_x3)*(r_x2-other.r_x3) + (r_y2-other.r_y3)*(r_y2-other.r_y3));
+    // NOTE: This assumes (even for right-to-left text) that other is
+    // to the right of this, as the calling code seems to do that.
 
-    return ((d0<eps) and (d1<eps));
+    // lower_right(this) : lower_left(other)
+    double dx0 = other.r_x0 - r_x1;
+    double dy0 = other.r_y0 - r_y1;
+    double d0 = std::sqrt(dx0 * dx0 + dy0 * dy0);
+
+    if (d0 >= eps)
+      return false;
+
+    // upper_right(this) : upper_left(other)
+    double dx1 = other.r_x3 - r_x2;
+    double dy1 = other.r_y3 - r_y2;
+    double d1 = std::sqrt(dx1 * dx1 + dy1 * dy1);
+
+    if (d1 >= eps)
+      return false;
+
+    return true;
   }
 
   bool pdf_resource<PAGE_CELL>::has_same_reading_orientation(pdf_resource<PAGE_CELL>& other)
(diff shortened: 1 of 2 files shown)


dhdaines:
Note that this implementation is obviously incorrect for text rotated outside (-90, 90) degrees.  I don't know if the bounding boxes used by docling-parse are normalized or not, and I don't know if the CTM is easily accessible from these classes (I've already wasted too many minutes of my life waiting for the C++ code to finish linking and wish to do no more) so I don't know the correct way to implement it for other rotations.

src/v2/pdf_resources/page_cell.h
```diff
@@ -262,10 +262,38 @@ namespace pdflib
...
-    double d0 = std::sqrt((r_x1-other.r_x0)*(r_x1-other.r_x0) + (r_y1-other.r_y0)*(r_y1-other.r_y0));
-    double d1 = std::sqrt((r_x2-other.r_x3)*(r_x2-other.r_x3) + (r_y2-other.r_y3)*(r_y2-other.r_y3));
+    // This assumes (even for right-to-left text) that other is to the
+    // right of this.  If two cells overlap then they are obviously
+    // adjacent, otherwise the right side corners of this must be
+    // within eps of the left side corners of other.
+
+    // Intersection of bounding rectangles (FIXME: Does not actually
+    // imply overlap for rotated cells, not immediately sure the
+    // correct and efficient way to compute that)
```
PeterStaar-IBM, line 272: Actually, if we are going to update this code, let's think this through how we deal with rotated cells.

dhdaines: Probably a good idea - is there documentation on what the rotated coordinates (`r_x0`, `r_y0`, etc, etc) mean exactly?  I assume that you take the four corners of the bounding box in user space and apply the CTM?  Are they normalized in any way (i.e. so `r_x0, r_y0` is always the bottom left point)?

dhdaines: In any case it's just the intersection of two convex polygons - and you don't actually need to find the resulting polygon, just its area (and even then you just need to know if the area is non-zero)

dhdaines: Presumably we can just use the algorithm from `shapely.intersects`: https://shapely.readthedocs.io/en/2.0.6/reference/shapely.intersects.html

dhdaines: Ah, that algorithm is actually in GEOS: https://github.com/libgeos/geos/blob/f1519c182497a99db8315ef78e0ae283b0469008/src/geom/Geometry.cpp#L291

dhdaines: likely it can be done by testing:
- if any corner point of one quadrilateral is inside the other
- if any line segment in one quadrilateral crosses the other

PeterStaar-IBM: @dhdaines Yes, currently the convention for the coordinates is,

```
      rx_4, ry_4        rx_3, ry_3
         +----------------+
         |                |
         |                |
         |                |
         +----------------+
      rx_0, ry_0        rx_1, ry_1
```
      
Let me look into the algos you send the links for. I do think we need to get this right, there are other example pdf's where we have similar problems. 
  
However, I do not want to break this example: [page](https://github.com/DS4SD/docling-parse/blob/main/docs/visualisations/table_of_contents_01.pdf.page_1.char.png)

dhdaines: thank you! I understand the function of `rx_*, ry_*` now.  Are `x0, y0, x1, y1` the bounding box of this rotated rectangle, or are they just the opposite corners of the user-space bounding box, after translation?  (if the latter, you may wish to consider this problem: https://github.com/pdfminer/pdfminer.six/issues/1067)

dhdaines: You may wish to use the approach I use in PLAYA to verify the behaviour with various possible types of transformation: https://github.com/dhdaines/playa/blob/ad15b4b4eefc529ebd6e079fcc8411e37466c911/tests/test_lazy_api.py#L104

PeterStaar-IBM: yes, so consider the (rx_*, ry_*) as coordinates of a polygon. (rx_0, ry_0) is the left bottom starting point of the writing and allows you to see the orientation of writing. The x/y coordinates are defined as 

x0 = min(rx_0, rx_1, rx_2, rx_3)
y0 = min(ry_0, ry_1, ry_2, ry_3)
x1 = max(rx_0, rx_1, rx_2, rx_3)
y1 = max(ry_0, ry_1, ry_2, ry_3)

Hence, they can not provide you the orientation of writing. 

Notice also that the coordinate origin is always (in pdf) bottom left corner of the page-image (as opposed to top left often the case with image libraries).

If we are going to solve this problem, we should not use the x0, y0, x1, y1 coordinates, but only the r*_* coordinates.

cau-git: @dhdaines could you let us know if you can address some of the open points raised in this discussion? I would love if this proceeds but I cannot dig deeper into the specifics at this time. Many thanks for your help!

dhdaines: > @dhdaines could you let us know if you can address some of the open points raised in this discussion? I would love if this proceeds but I cannot dig deeper into the specifics at this time. Many thanks for your help!

yes absolutely!  I haven't had time in the last 2 weeks but I'll take a look at it today.  It may be the case that we can just fix the underlying issue with overlapping glyph bboxes - possibly the charspace or wordspace are being applied incorrectly in the case where each glyph is a separate text cell.

dhdaines: See updates in #99 - the underlying problem is that `docling-parse` disregards the coordinate space in Type3 fonts.  So this whole MR is mostly irrelevant, if we can fix that problem (which should be A Simple Matter Of Matrix Multiplication™)

dhdaines:
Here is a Python implementation of convex polygon intersection, C++ is left as an exercise to the reader:
```python
from typing import Iterator, Sequence, Tuple
Point = Tuple[float, float]

def edges(poly: Sequence[Point]) -> Iterator[Tuple[Point, Point]]:
    """Iterate over edges of a polygon."""
    j = len(poly) - 1
    for i in range(len(poly)):
        yield poly[i], poly[j]
        j = i

def pnpoly(point: Point, poly: Sequence[Point]) -> bool:
    """Use even-odd rule to determine point-in-polygon."""
    x, y = point
    inside = False
    for (xi, yi), (xj, yj) in edges(poly):
        if (yi > y) is not (yj > y):
            if x < (xj - xi) * (y - yi) / (yj - yi) + xi:
                inside = not inside
    return inside
    
def lines_intersect(p1: Point, p2: Point, p3: Point, p4: Point) -> bool:
    """Determine if two lines intersect."""
    if p1 == p2 or p3 == p4:  # zero-length lines
        return False
    x1, y1, x2, y2, x3, y3, x4, y4 = *p1, *p2, *p3, *p4
    denom = (y4 - y3) * (x2 - x1) - (x4 - x3) * (y2 - y1)
    if denom == 0:  # parallel lines
        return False
    ua = ((x4 - x3) * (y1 - y3) - (y4 - y3) * (x1 - x3)) / denom
    ub = ((x2 - x1) * (y1 - y3) - (y2 - y1) * (x1 - x3)) / denom
    if ua < 0 or ua > 1 or ub < 0 or ub > 1:
        return False
    return True

def polys_collide(a: Sequence[Point], b: Sequence[Point]) -> bool:
    """Determine if two convex polygons overlap."""
    for ae in edges(a):
        for be in edges(b):
            if lines_intersect(*ae, *be):
                return True
    return False

def polys_intersect(a: Sequence[Point], b: Sequence[Point]) -> bool:
    """Determine if two convex polygons intersect (including the case
    where one is contained entirely in the other)."""
    if polys_collide(a, b):
        return True
    for point in a:
        if pnpoly(point, b):
            return True
    for point in b:
        if pnpoly(point, a):
            return True
    return False
```

dhdaines:
So, I've implemented the overlap detection, but it doesn't actually fix the issue, because the character bboxes are still wrong for Type3 fonts.  But also it causes some problems for instance in https://github.com/docling-project/docling-parse/blob/main/docs/visualisations/table_of_contents_01.pdf.page_2.word.png where the characters in the figures legitimately have overlapping bboxes.

I'm closing this PR for the moment, though you may wish to keep the code for future use.  I think a fix for the underlying Type3 font problem should be fairly easy so I'll make a new PR for that.

PR #113, opened 2025-03-27
fix: Use FontMatrix to scale Type3 font metrics

### LINKED ISSUE #99: Wrong text output

dhdaines:
`get_unit` didn't do anything useful, but also it isn't sufficient to just scale the font size in the case of Type3 fonts - the font matrix can (and often does) flip the axes of the glyph space as well as rescaling it, so we need to take this into account when calculating ascent and descent.

Fixes #99


PeterStaar-IBM:
@dhdaines Awesome work, I can only say: big thanks for this fix and the additional test case!

A few little comments:

1. would it be possible to avoid `==` and `!=` operations of the components of the font-matrix. It might be nice to set a boolean variable (maybe `is_font_type_3`?) in the function `init_font_matrix()`. I am just a bit worried that on different systems the `!=` on doubles might a buggy.
2. Can you always put `{}` around what follows the if-statements
3. Could you also check that the `result` vector from the font-matrix is of length 6 (line: 891)?

Again big thanks and well done!

src/v2/pdf_resources/page_font.h
```diff
@@ -891,6 +879,37 @@ namespace pdflib
...
+  {
+    LOG_S(INFO) << __FUNCTION__;// << "\t" << json_font.dump(2);
+
+    std::vector<std::string> keys_0 = {"/FontMatrix"};
+
+    if(utils::json::has(keys_0, json_font))
+      {
+        //assert(subtype==TYPE_3);
+        auto result = utils::json::get(keys_0, json_font);
+
```
PeterStaar-IBM, line 892: Let's check that result has a size of 6 before we continue

src/v2/pdf_resources/page_font.h
```diff
@@ -891,6 +879,37 @@ namespace pdflib
...
+          {
+            font_matrix[d] = result[d].get<double>();
+          }
+      }
+    else
+      {
+        LOG_S(INFO) << "using default font-matrix";
+        font_matrix = {0.001, 0, 0, 0.001, 0, 0};
+      }
+
```
PeterStaar-IBM, line 903: Let's set an internal boolean flag here which analyses the font-matrix and avoids us to do `!=` downstream.

src/v2/pdf_resources/page_font.h
```diff
@@ -1166,7 +1185,15 @@ namespace pdflib
         {
           LOG_S(WARNING) << "'xheight' was not explicitely defined ...";
         }
-    }    
+    }
+
+    if (font_matrix[3] != 0.001)
```
PeterStaar-IBM, line 1190: replace this inequality with the newly introduced boolean flag.

src/v2/pdf_resources/page_font.h
```diff
@@ -1305,6 +1332,8 @@ namespace pdflib
 	  }
 	
         numb_to_widths[ind] = values[cnt++];
+        if (font_matrix[0] != 0.001)
```
PeterStaar-IBM, line 1335: replace this inequality with the newly introduced boolean flag.

src/v2/pdf_resources/page_font.h
```diff
@@ -1305,6 +1332,8 @@ namespace pdflib
...
+          numb_to_widths[ind] *= font_matrix[0] * 1000.0;
```
PeterStaar-IBM, line 1336: Add `{ ... }` around

src/v2/pdf_resources/page_font.h
```diff
@@ -1383,6 +1412,8 @@ namespace pdflib
               {
 		//LOG_S(WARNING) << "\t" << id << " -> " << w;
                 numb_to_widths[id] = w;
+                if (font_matrix[0] != 0.001)
```
PeterStaar-IBM, line 1415: Replace this inequality with the newly introduced boolean flag.

src/v2/pdf_resources/page_font.h
```diff
@@ -1404,6 +1435,8 @@ namespace pdflib
 		//LOG_S(WARNING) << "\t" << beg+k  << " -> " << w[k];
 
                 numb_to_widths[beg+k] = w[k];
+                if (font_matrix[0] != 0.001)
```
PeterStaar-IBM, line 1438: Replace this inequality with the newly introduced boolean flag.

src/v2/pdf_resources/page_font.h
```diff
@@ -1383,6 +1412,8 @@ namespace pdflib
               {
 		//LOG_S(WARNING) << "\t" << id << " -> " << w;
                 numb_to_widths[id] = w;
+                if (font_matrix[0] != 0.001)
+                  numb_to_widths[id] *= font_matrix[0] * 1000.0;
```
PeterStaar-IBM, line 1416: Add { ... } around

src/v2/pdf_resources/page_font.h
```diff
@@ -1404,6 +1435,8 @@ namespace pdflib
 		//LOG_S(WARNING) << "\t" << beg+k  << " -> " << w[k];
 
                 numb_to_widths[beg+k] = w[k];
+                if (font_matrix[0] != 0.001)
+                  numb_to_widths[beg+k] *= font_matrix[0] * 1000.0;
```
PeterStaar-IBM, line 1439: Add { ... } around

PeterStaar-IBM (changes requested):

dhdaines:
> 1. would it be possible to avoid `==` and `!=` operations of the components of the font-matrix. It might be nice to set a boolean variable (maybe `is_font_type_3`?) in the function `init_font_matrix()`. I am just a bit worried that on different systems the `!=` on doubles might a buggy.

Ah, indeed - I wasn't worried about it because it is checking for the default case where we have explicitly set the value to the constant 0.001 elsewhere, but you're right, that isn't a comparison that is reliably true on floating-point values.

> 2. Can you always put `{}` around what follows the if-statements
> 3. Could you also check that the `result` vector from the font-matrix is of length 6 (line: 891)?

Yes, of course ... I'm not familiar with the JSON library in question or its API...

src/v2/pdf_resources/page_font.h
dhdaines: Sure, I will also add similar checks in `init_font_bbox`

PeterStaar-IBM:
> > 1. would it be possible to avoid `==` and `!=` operations of the components of the font-matrix. It might be nice to set a boolean variable (maybe `is_font_type_3`?) in the function `init_font_matrix()`. I am just a bit worried that on different systems the `!=` on doubles might a buggy.
> 
> Ah, indeed - I wasn't worried about it because it is checking for the default case where we have explicitly set the value to the constant 0.001 elsewhere, but you're right, that isn't a comparison that is reliably true on floating-point values.

Yes, I have been burnt a few too many times!

> > 2. Can you always put `{}` around what follows the if-statements
> > 3. Could you also check that the `result` vector from the font-matrix is of length 6 (line: 891)?
> 
> Yes, of course ... I'm not familiar with the JSON library in question or its API...

No worries! For the array, just use,

```
if (j.is_array() && j.size() == 6) {
    // it's an array and has size 6
} 
else {
   // Let's do some error message here
}
```

=> Again, big thanks and looking forward merging this in!

dhdaines:
Should be ready now!

PeterStaar-IBM:
Very nice, I will take it from here to get it all merged. Let's verify all the tests pass and we should be good to go!

dolfim-ibm:
For reference, here is the usage of this PR in the Docling tests: https://github.com/docling-project/docling/pull/1326 (it compares against the docling-parse v4 test updates).

✅  The only difference is tiny diff in a bbox.

dolfim-ibm (approved):
lgtm

PeterStaar-IBM (approved):
awesome work, deserves 🎖️ !

PR #117, opened 2025-04-09
Update pdf_parser.py

cahoover:
feat: added font_size to `PdfTextCell` object

My project needs a font size (if available) within the pdf spans. It appears the C++ already contains this information, so we added it to the python class.

The change ensures that if the `font-size` key exists in the data provided by the C++ layer, its value is read and passed during the creation of the `PdfTextCell` object. 

Note we also changed docling-core/docling_core/types/doc/page.py to include an optional `font_size` field. This is reflected in a separate pull request.


cahoover:
Looks like this issue is deeper in the C++

PR #124, opened 2025-06-04
fix: filenames with unicode chars on Windows

dolfim-ibm:
Resolves https://github.com/docling-project/docling/issues/1218


PeterStaar-IBM (approved):
lgtm!

PR #135, opened 2025-06-18
feat: switch to uv

rmdg88:
- switch from poetry to uv
- updated runners image
- updated externals
- updated docs


dolfim-ibm (approved):
lgtm

PR #171, opened 2025-10-09
fix: Make build and dev dependencies optional

### LINKED ISSUE #170: Error in pyproject.toml leads to cmake, etc. being included in the resulting package
There's a minor error in pyproject.toml -- defining `build`, `dev`, and `perf-tests` as dependency groups packages all of them into the final build. (At least in nixpkgs; but this is likely to be happening elsewhere.) The correct approach is to define these as optional dependencies.

sarahec:
This fixes #170

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
-[dependency-groups]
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM:
already taken care of by other PR, but thanks!

Issue #140, opened 2025-06-23
Docling Conversion outputs ASCII codes instead of text


Saidgurbuz:
I encountered an issue while using the Docling CLI (version 2.36.1) to convert a PDF to Markdown. I ran the conversion on the following PDF:
` https://intra.ece.ucr.edu/~rlake/Whitesides_writing_res_paper.pdf`
with the following command:
```bash
docling https://intra.ece.ucr.edu/~rlake/Whitesides_writing_res_paper.pdf
```

The generated Markdown file contains encoded strings like:
```
/C87/C104/C101/C110
```
These appear to be ASCII character codes. When I manually convert these codes to characters, the result is actually a correct Markdown version of the PDF’s content.

PR #188, opened 2026-01-09
fix: respect declared font encoding over base font mapping

### LINKED ISSUE #187: MacRomanEncoding ignored when font matches known base font (e.g., Arial)
### Bug

I traced this bug using Claude Code.

When a PDF font declares `/Encoding /MacRomanEncoding` but matches a known base font (e.g., Arial), the code uses the base font's built-in WinAnsi character mapping instead of respecting the declared encoding. This causes character 0xa1 to be extracted as `¡` (WinAnsi) instead of `°` (MacRoman).

**Example:** `LPG Vapor Pressure @100°F` → `LPG Vapor Pressure @100¡F`

**Root cause:** In `src/v2/pdf_resources/page_font.h`, function `get_correct_character()` at lines 480-498, when a font matches a "corresponding font" (like ArialMT), the code uses `fm.to_utf8(c)` from the base font's `.afm` file without checking if the PDF font declares a different encoding.

Arial.afm has: `C 161 ; WX 333 ; N exclamdown` (WinAnsi mapping)
But MacRoman encoding has: `a1;degree` → `°`

**Fix:** Before using the base font's mapping, check if the font declares a specific encoding:

```cpp
// If font declares a specific encoding (MacRoman, WinAnsi, etc.),
// use that encoding instead of base font's built-in mapping
if(encoding == MACROMAN || encoding == MACEXPERT || encoding == WINANSI || encoding == STANDARD)
  {
    return get_character_from_encoding(c);
  }
else if(fm.has(c))
  {
    return fm.to_utf8(c);
  }
```

I've tested this fix locally and confirmed it resolves the issue.

### Steps to reproduce

1. Process a PDF with a subsetted Arial font declaring MacRoman encoding:
    ```
    /BaseFont /PXAAAB+ArialMT
    /Encoding /MacRomanEncoding
    ```
2. Extract text containing character code 0xa1
3. Observe `¡` instead of `°`

### Test file

See attached `macroman_encoding_bug_demo.pdf` - a minimal 1.7KB PDF that reproduces the issue.

### Docling version

2.65.0 (docling-parse 2.1.2)

### Python version

3.12

silverl:
> This PR was researched and coded by [Claude Code](https://docs.anthropic.com/en/docs/claude-code).

## Summary

Fixes #187 - MacRomanEncoding fonts incorrectly decoded using WinAnsi mappings.

When a PDF embeds a subsetted font (e.g., `PXAAAB+ArialMT`) with a declared encoding like `/Encoding /MacRomanEncoding`, the parser was ignoring this declaration and using character mappings from the matched base font's `.afm` file instead.

**Root cause:** In `get_correct_character()`, when `bfonts.has_corresponding_font()` matches a known font, the code used `fm.to_utf8(c)` which applies the base font's internal mapping (from the `.afm` file). For Arial, this is effectively WinAnsi encoding, causing byte `0xa1` to become `¡` (exclamdown) instead of `°` (degree symbol) for MacRoman-encoded fonts.

## The Fix

**Part 1 (main fix):** In `get_correct_character()`, check if the font declares a specific encoding (MacRoman, WinAnsi, MacExpert, or Standard) before falling back to base font mappings. If a declared encoding exists, use the encoding table instead.

**Part 2 (defensive):** In `init_encoding()`, extract `/BaseEncoding` from encoding dictionary objects. Added `.is_string()` type check for safety.

## Test Impact

Some regression tests will fail because their ground truth files were generated with the buggy code. Affected PDFs (like `form_fields.pdf`) contain MacRoman-encoded fonts that now decode correctly. To update:

1. Set `GENERATE = True` in `test_parse.py` and `test_parse_v2.py`
2. Run tests to regenerate ground truth
3. Set `GENERATE = False`
4. Commit updated ground truth files

A minimal test PDF (`macroman_encoding_bug_demo.pdf`) demonstrating this bug is attached to issue #187.

## Example

PDF with `/Encoding /MacRomanEncoding` and byte `0xa1`:
- **Before:** `¡` (WinAnsi mapping from Arial.afm)
- **After:** `°` (correct MacRoman mapping)

## Why This Is Correct

Per PDF specification (Section 5.5.5), a declared encoding overrides any default encoding associated with the font. The fix ensures the parser respects this hierarchy:

1. ToUnicode CMap (if present) - unchanged
2. Differences array (if present) - unchanged
3. **Declared encoding (MacRoman/WinAnsi/etc.)** - **NEW: takes precedence**
4. Base font mapping - fallback for UNKNOWN encoding only


PeterStaar-IBM:
@silverl Can you provide also a test-case? Add one (or more) single page pdf's in which this occurs. I always like to have regression tests for these font updates.

PeterStaar-IBM:
would be interesting to see if it also solves this issue: https://github.com/docling-project/docling-parse/issues/140

PeterStaar-IBM:
@silverl Thanks for this PR, it gave me the good direction and I have now 

1. updated your code slightly here (https://github.com/docling-project/docling-parse/pull/193). It was missing some small part to make sure we dont break other regression tests
2. added the test pdf as a regression test

silverl:
That's great news, thanks @PeterStaar-IBM. I checked on pypi, but no build has appeared yet. It looks like the GitHub action might have stalled out. The 4.7.2 took 1h15m, but the current build is still going after 6 hours.

silverl:
Confirmed fixed! I've synced up with pypi and docling-parse 4.7.3 and it's working great.

PR #204, opened 2026-01-29
fix: recursively traverse parent chain for inherited MediaBox

### LINKED ISSUE #175: RuntimeError: could not find the page-dimensions on a PDF
Link to the original PDF: http://europa.eu/eur-lex/lex/LexUriServ/LexUriServ.do?uri=OJ:L:2002:032:0022:0037:EN:PDF

Code to reproduce:
```python
from docling_parse.pdf_parser import DoclingPdfParser

parser = DoclingPdfParser()
pdf_doc = parser.load("test.pdf")
pdf_doc.get_page(1)
```

Docling Parse version: 4.7.0

emerose:
## Summary

PDF specification (sec 7.7.3.4, p 80) allows `/MediaBox` to be inherited from any ancestor in the page tree, not just the immediate parent. The previous fix in v4.7.1 only checked one level of parent, which failed for PDFs where `/MediaBox` is defined at the grandparent level or higher (common in multi-level page trees).

## Problem

Certain valid PDFs fail with:
```
RuntimeError: could not find the page-dimensions: {"/Type": "/Page"}
```

These PDFs have `/MediaBox` defined at the root of the page tree (e.g., grandparent), not on individual pages or their immediate parents.

### Example PDF Structure (EU-lex document)
```
Page (xref 1)
  └─ Parent: Pages node (no /MediaBox)
       └─ Parent: Root Pages node (/MediaBox [0 0 595 842]) ← Here!
```

## Solution

Recursively traverse the parent chain (up to 10 levels to prevent infinite loops in malformed PDFs) until `/MediaBox` is found or the root is reached.

```cpp
for(int depth = 0; depth < 10 && current.hasKey("/Parent"); depth++)
{
    QPDFObjectHandle parent = current.getKey("/Parent");
    if(parent.hasKey("/MediaBox"))
    {
        // Extract MediaBox from this ancestor
        ...
    }
    current = parent;
}
```

## Testing

- All existing tests pass (16/16)
- Added test case: `deep-mediabox-inheritance.pdf` (EU-lex document with MediaBox at grandparent level)
- Tested with multiple real-world PDFs that previously failed

## Fixes

Fixes #175

Signed-off-by: Sam Quigley <sq@emerose.com>

### DIFF
--- a/src/parse/pdf_resources/page_dimension.h
+++ b/src/parse/pdf_resources/page_dimension.h
@@ -245,27 +245,46 @@ namespace pdflib
       }
 
     if(json_resources.count("/MediaBox"))
-      {        
+      {
         for(int d=0; d<4; d++)
           {
             media_bbox[d] = json_resources["/MediaBox"][d].get<double>();
           }
       }
-    // it might inherit the media-bbox from the parent document (sec 7.7.3.4, p 80)
-    else if(qpdf_resources.hasKey("/Parent") and qpdf_resources.getKey("/Parent").hasKey("/MediaBox"))
-      {
-	QPDFObjectHandle qpdf_bbox = qpdf_resources.getKey("/Parent").getKey("/MediaBox"); 
-	nlohmann::json   json_bbox = to_json(qpdf_bbox);
-
-	//LOG_S(WARNING) << "inherited bbox: " << json_bbox.dump(2);	
-	for(int d=0; d<4; d++)
-          {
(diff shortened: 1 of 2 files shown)


PeterStaar-IBM:
@emerose Love it! Thanks for the code + the test-case.

You do have to add the ground-truth files too (they get created otherwise)

PeterStaar-IBM (approved):
wonderful!

PR #230, opened 2026-02-23
fix: merge mixed-font symbols into line cells

serge-medvedev:
Use the order-independent line merge path when building textline cells and disable same-font enforcement for line-cell merges. This prevents fallback-font symbols (for example arrows) from being emitted as standalone line cells when PDF content-stream order differs from visual order.

Also make v2 line merging handle reverse adjacency symmetrically and add a targeted synthetic regression fixture/test that asserts the exact textline-cell output.


PeterStaar-IBM (changes requested):

serge-medvedev:
@PeterStaar-IBM , confirmed.

PeterStaar-IBM:
> @PeterStaar-IBM , confirmed.

then I close this PR. ok?

PR #277, opened 2026-05-27
feat: add materialize_bitmap_bytes flag to skip bitmap byte extraction

cau-git:
## Summary

- Add `materialize_bitmap_bytes: bool = True` to `DecodePageConfig` (C++ struct, consumed Python-side only). When `False`, `bitmap_resources` is populated with geometry but byte extraction is skipped — `BitmapResource.image` is `None`, `mode` is `PLACEHOLDER`.
- Widen `PdfDocument._pages` cache key from `int` → `tuple[int, bool]` to prevent a placeholder-cached page from satisfying a materialized request.
- `PageParseResult` receives `decode_config` at construction; `get_page()` reads `config.materialize_bitmap_bytes` — consistent with the non-threaded path, no separate argument.
- Add `__copy__` / `__deepcopy__` to the `DecodePageConfig` pybind binding; remove the hand-rolled `_copy_decode_config` function. New C++ fields are automatically covered by `copy.copy()`.


PeterStaar-IBM (approved):
lgtm!

PR #279, opened 2026-06-09
feat: prevent O(N²) XObject reloading in do_form() causing OOM on complex PDFs

seongmincho315:
## Summary

PDFs containing many Form XObjects that share a large resource dictionary (e.g. a scatter plot with 2,296 data points, each referencing 2,301 sibling XObjects) caused an OOM kill during parsing.

**Root cause:** In `do_form()` (`src/parse/pdf_decoders/stream.h`), `set()` was unconditionally called for fonts, graphics states, and XObjects on every invocation. When `do_form()` is called N times and each call reloads all N resources, the result is O(N²) memory allocations — 2,296 × 2,301 ≈ 5.3 million redundant reloads in the reported case.

**Fix:** Before calling `set()`, check whether all keys in the child resource dictionary are already present in the parent chain. If so, skip the `set()` call. Resources inherited from a parent Form XObject are already available, so re-registering them on every `do_form()` call is unnecessary.

## Test case

- PDF: [arXiv:2508.13113](https://arxiv.org/abs/2508.13113) (page 2 — scatter plot with 2,296 data points)
- Before fix: OOM kill at ~200 s
- After fix: page parsed successfully in ~181 s

Related issue: https://github.com/docling-project/docling/issues/2109

## Checklist

- [x] Signed-off-by included (DCO)
- [x] Single file changed (`src/parse/pdf_decoders/stream.h`)
- [x] No behavior change for PDFs without repeated Form XObject resource dictionaries

### DIFF
--- a/src/parse/pdf_decoders/stream.h
+++ b/src/parse/pdf_decoders/stream.h
@@ -1,5 +1,6 @@
 //-*-C++-*-
 
+
 #ifndef PDF_STREAM_DECODER_H
 #define PDF_STREAM_DECODER_H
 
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM (approved):
lgtm!

PeterStaar-IBM:
@seongmincho315 #281  solves the problem from first principles (the `to_json` was the culprit). Unfortunately, the current proposal breaks some of the unit-tests.

PR #282, opened 2026-06-25
feat: LRU cache eviction for PDF page decoders

dpantaleoni:
## Description

This PR adds a Least Recently Used cache eviction policy for pdf page decoders, preventing unbounded memory growth when processing many large PDF documents in succession. This resolves the bug described in this [comment](https://github.com/docling-project/docling-serve/issues/366#issuecomment-4728834016) on [docling-serve issue #366](https://github.com/docling-project/docling-serve/issues/366).

## Fix

Replacing unbounded page decoder accumulation with an LRU-bounded cache of max 16 live page decoders eliminated the observed steady RSS growth when converting many large PDFs in succession.

Before:
before-res_chronological_rss_lines

After:
after-res_chronological_rss_lines

## Testing

The script below was ran within a docling dev environment with the locally edited docling-parse dependency installed. A folder of real-world, diverse pdfs of varying sizes ~1 MB - ~20 MB was used. This was done on a Macbook pro m4 36 GB.
[check_mem.py](https://github.com/user-attachments/files/29351744/check_mem.py)

### DIFF
--- a/src/parse/pdf_decoders/document.h
+++ b/src/parse/pdf_decoders/document.h
@@ -89,6 +89,24 @@ namespace pdflib
 
     // New: Persistent page decoders for typed API
     std::map<int, page_decoder_ptr> page_decoders;
+    std::list<int> page_access_order;
+    size_t max_cached_pages = 16;
+
+    void update_page_access(int page_number) {
+      // Remove from current position if exists
+      page_access_order.remove(page_number);
+      // Add to back (most recently used)
+      page_access_order.push_back(page_number);
+    }
+    
+    void evict_lru_page() {
+        if (!page_access_order.empty()) {
+            int lru_page = page_access_order.front();
+            page_access_order.pop_front();
+            page_decoders.erase(lru_page);
+            LOG_S(INFO) << "Evicted LRU page " << lru_page << " from cache";
+        }
+      }
   };
 
   pdf_decoder<DOCUMENT>::pdf_decoder():
@@ -450,6 +468,9 @@ namespace pdflib
 
     // Store in cache
     page_decoders[page_number] = page_decoder;
+    update_page_access(page_number);
+    // LRU eviction if cache exceeds limit
+    if(page_decoders.size() > max_cached_pages) evict_lru_page(); 
 
     std::stringstream ss;
     ss << pdf_timings::PREFIX_DECODE_PAGE << page_number;
@@ -463,6 +484,7 @@ namespace pdflib
     if(page_decoders.count(page_number) > 0)
       {
         page_decoders.erase(page_number);
+        page_access_order.remove(page_number); 
         LOG_S(INFO) << "unloaded page decoder for page: " << page_number;
       }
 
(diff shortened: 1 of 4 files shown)


src/parse/pdf_decoders/document.h
```diff
@@ -89,6 +89,24 @@ namespace pdflib
 
     // New: Persistent page decoders for typed API
     std::map<int, page_decoder_ptr> page_decoders;
+    std::list<int> page_access_order;
+    size_t max_cached_pages = 16;
```
PeterStaar-IBM: this needs to be a configurable parameter for sure.

dpantaleoni: After making max_cached_pages configurable and setting default to -1 (no limit, to follow set conventions), I believe both docling and docling-serve will need a few changes to allow the user to set max_cached_pages to a limit. Would this be okay?

PeterStaar-IBM: @dpantaleoni we need to set sensible default values. We do not necessarily need to propagate these variables to docling/docling-serve.

PeterStaar-IBM:
@dpantaleoni Also have a look at this one: https://github.com/docling-project/docling-parse/pull/287

I believe this is a better approach, since we limit the total number of page-decoders, not just the results per-document.

dpantaleoni:
@PeterStaar-IBM Yeah this is definitely a better approach; global limiting is more what I wanted to do. I should have dove deeper into the parsing pipeline before settling on a solution! Would you like me to continue off the work in your PR (adding config and memray testing)?

dpantaleoni:
@PeterStaar-IBM Also sorry for the slow progress, the benchmarking runs are slow on my machine. I've also been trying to decipher the memray results and haven't been satisfied with what I've seen.

PeterStaar-IBM:
@dpantaleoni I think this PR finally solves it: https://github.com/docling-project/docling-parse/pull/287. Can you try to reproduce the plots above in your setup with #287 . If it is the same or better, we can close this one.

PeterStaar-IBM:
everything seems to have been fixed.

dpantaleoni:
@PeterStaar-IBM This is the equivalent test results with the #287 fix only. ~5 MB/conv growth on my macbook. so much better than ~20 MB/conv.
after_globalfix_results2_chronological_rss_lines

And this is with the #287 fix and #282 fix (also limiting per pdf page decoders to 16). Also ~5 MB/conv so basically the same.
global_perpdf2_chronological_rss_lines

I ran both of these twice to confirm and got similar results. I'd be curious to see if @sergeyshmakov sees improvement as well.

PeterStaar-IBM:
> @PeterStaar-IBM This is the equivalent test results with the #287 fix only. ~5 MB/conv growth on my macbook. so much better than ~20 MB/conv. after_globalfix_results2_chronological_rss_lines
> 
> And this is with the #287 fix and #282 fix (also limiting per pdf page decoders to 16). Also ~5 MB/conv so basically the same. global_perpdf2_chronological_rss_lines
> 
> I ran both of these twice to confirm and got similar results. I'd be curious to see if @sergeyshmakov sees improvement as well.

awesome!

PR #288, opened 2026-07-01
feat: build wheel for windows arm64

dolfim-ibm:


PeterStaar-IBM (dismissed):
lgtm!

PeterStaar-IBM:
For the Windows Arm64, we needed to update .github/workflows/wheels.yml and cmake/extlib_jpeg.cmake

  - Root cause: on the windows-11-arm runner, -A ARM64 correctly sets the MSBuild target platform, but CMAKE_SYSTEM_PROCESSOR still resolves to AMD64 (host-reported arch quirk). libjpeg-turbo's CMake keys its SIMD backend choice off CMAKE_SYSTEM_PROCESSOR, so it built x86_64 NASM SIMD objects that then failed to link against the ARM64 target (LNK1112).
  - Fix: force CMAKE_SYSTEM_PROCESSOR=ARM64 in the win_arm64 CI job's CMAKE_ARGS, and forward that same variable into extlib_jpeg.cmake's ExternalProject_Add (which doesn't auto-inherit plain -D cache vars from the parent configure — only generator/platform settings propagate automatically).

PeterStaar-IBM (approved):
lgtm!

Issue #180, opened 2025-11-14
[Bug] Incorrect Character Width Calculation with Td Commands


flexorRegev:
[test_courier_td_bug.pdf](https://github.com/user-attachments/files/23555003/test_courier_td_bug.pdf)

## Summary

docling-parse incorrectly calculates character widths when parsing PDFs that use `Td` (text displacement) commands with long text strings, resulting in word bounding boxes that are **consistently 20% too small** (5pt per character instead of the correct 6pt for Courier 10pt font).

## Environment

- **docling-parse version**: Latest (tested with current main branch)
- **Python version**: 3.12
- **Operating System**: Linux

## Problem Description

When parsing PDFs that use `Td` commands followed by text strings containing spaces, docling-parse produces bounding boxes that are systematically too narrow. The issue appears to be caused by using a hardcoded character width assumption (5pt) rather than reading actual font metrics from the PDF.

### Expected Behavior

For a Courier 10pt font, each character should be **6.00 points wide** (Courier is a monospace font with 600 units per character in standard PDF metrics, which equals 6pt at 10pt font size).

### Actual Behavior

docling-parse calculates character widths as **5.00 points**, resulting in:
- Bounding boxes that are 5/6 (≈83.3%) of their correct width
- Word positions that drift progressively left as they appear later in the text string
- A consistent 1.2× ratio between correct width and docling-parse width

### Minimal Test PDF

I've attached `test_courier_td_bug.pdf` which reproduces this issue. The content stream is:

```pdf
BT
/F001 10 Tf
72 720 Td
( ABC12345                           COMPANY NAME INC.                          DOCUMENT TYPE) Tj
0 -11 Td
(ID=99999 REF1234567890123) Tj
0 -11 Td
(CUSTOMER : GENERIC STORE NAME                     LOCATION FROM : WESTERN) Tj
ET
```

With minimal font descriptor:
```pdf
<<
  /Type /Font
  /Subtype /Type1
  /BaseFont /Courier
>>
```

**Key characteristics that trigger the bug:**
- Standard Type1 Courier font with **no explicit width table** (relies on base font metrics)
- Uses `Td` command to set initial text position (e.g., `72 720 Td`)
- Text shown with `Tj` command containing **single long string (90+ characters)** with many embedded spaces
- Each `Td` command followed by `Tj` with text

The critical element is the **long string with embedded spaces** - docling-parse must walk through character-by-character to determine where each word begins and ends, using character widths to calculate positions.

## Impact

[...]

PR #294, opened 2026-07-05
fix: resolve core-14 font aliases to their standard metrics

pablopupo:
Standard Type1 fonts can be referenced by a common family alias like Arial or Times New Roman and legally ship without a /Widths table. docling-parse only found the bundled metrics when a canonical PostScript name (Helvetica, Times-Roman and so on) was a substring of the font name, so those aliases matched nothing and every glyph fell back to a flat 500 units/em. At 10pt that advances each character 5.00pt instead of its real width, which leaves word bounding boxes off and drifting further along the line the longer the string. On a 92 character Arial line the last glyph lands about 71pt away from where it should.

This is the remaining piece of #180. The canonical base-14 names from that report already resolve on current main (Courier gives the correct 6.00pt, Helvetica and Times-Roman give their real varying widths), so the change binds the Arial, Times New Roman and Courier New aliases to the Core-14 metrics that PDF viewers substitute. Arial now matches Helvetica glyph for glyph (A 6.67, W 9.44, i 2.22, m 8.33 at 10pt) instead of a flat 5.00. Fonts with an explicit /Widths table and names that already contain a canonical family keep their existing behavior, and a bundled real metric like ArialMT still wins over the alias. No new font data is added, the aliases reuse the Core-14 AFM files already shipped in pdf_resources.

Added tests/test_standard_font_widths.py, which builds its minimal PDFs in memory so it needs no external data. It covers Courier at 6.00pt, a proportional font with varying non-500 widths, the alias resolution for Arial and Times, and an explicit /Widths table taking precedence. The full test suite passes locally (81 passed).

### DIFF
--- a/src/parse/pdf_resources/page_font/base_fonts.h
+++ b/src/parse/pdf_resources/page_font/base_fonts.h
@@ -42,6 +42,8 @@ namespace pdflib
 
     std::string read_fontname(std::string filename);
 
+    void register_standard_aliases();
+
   private:
 
     bool initialized;
@@ -236,6 +238,8 @@ namespace pdflib
 	  }
       }
 
+    register_standard_aliases();
+
     initialized = true;
   }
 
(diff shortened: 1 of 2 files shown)


PeterStaar-IBM:
@pablopupo I have resolved already quite a few issues regarding this with the latest renderer. Can you attach here a PDF that does not work on main but that this PR solves?

pablopupo:
This is a small PDF that still breaks on current main, for example. It is one page with the same line of text three times, once each in Arial, TimesNewRoman and Courier New, none of them carrying a /Widths table. On main every glyph in the Arial and Times lines advances a flat 6.0pt at 12pt, so both lines come out 210.0pt wide. With this PR cherry-picked onto main (it applies cleanly) the same lines measure 162.64pt and 165.02pt, which matches the bundled Helvetica and Times AFM widths glyph for glyph, from 2.66pt for i and l up to 11.33pt for W. The CourierNew line is 252.0pt in both builds because courier already matches as a substring of the name, so what is left after your renderer work is the Arial and Times style aliases falling through to the 500 fallback.

https://github.com/pablopupo/docling-parse/raw/repro-294/core14-alias-no-widths.pdf

PeterStaar-IBM:
**Before**

Screenshot 2026-07-11 at 07 08 29

**after**

blend2d_renderer_preview

**Groundtruth**

Screenshot 2026-07-11 at 07 08 55

@pablopupo I am not 100% convinced we solved the entire problem. Definitely going in the right direction

PeterStaar-IBM:
OK, I fixed the rendering here: https://github.com/docling-project/docling-parse/pull/298

pablopupo:
Thanks for jumping on the renderer half so quickly! That matches what your screenshots showed, the advances were already right and the preview just needed the face resolution. I built #298 locally and re-ran my repro, the geometry comes out byte for byte identical to this branch (162.64 / 165.02 / 252.0, matching the AFMs) and all three lines paint with the right faces now, even with the MS fonts hidden so the alias fallback actually gets exercised.

Since #298 includes my commit unchanged, would you be open to merging this one first and then rebasing #298 so it's just the renderer commit on top? That keeps the authorship clean in the history, and it would also be my first PR in this repo. If that's a pain to sequence, no worries at all! 

Thank you!

PeterStaar-IBM:
@pablopupo Yes, I will merge this first (will fix the merge conflicts in case), I understand you ant credit, which you deserve.

One small ask: can you make an extended pdf (similar to https://github.com/pablopupo/docling-parse/raw/repro-294/core14-alias-no-widths.pdf) that has "all" the different fonts,

```
        {"arial",                    "helvetica"},
        {"arial-bold",               "helvetica-bold"},
        {"arial-italic",             "helvetica-oblique"},
        {"arial-bolditalic",         "helvetica-boldoblique"},
        {"couriernew",               "courier"},
        {"couriernew-bold",          "courier-bold"},
        {"couriernew-italic",        "courier-oblique"},
        {"couriernew-bolditalic",    "courier-boldoblique"},
        {"times",                    "times-roman"},
        {"timesnewroman",            "times-roman"},
        {"timesnewroman-bold",       "times-bold"},
        {"timesnewroman-italic",     "times-italic"},
        {"timesnewroman-bolditalic", "times-bolditalic"},
   ```
   
   and attach it in a comment on #298 ? I want to include that into our regression tests.

PeterStaar-IBM (approved):
looks good (and thanks for flagging this issue!)

PR #295, opened 2026-07-09
fix: arabic pdf-rendering

PeterStaar-IBM:
**Before**

right_to_left_03 pdf page_no_1 full_page

**After**

blend2d_renderer_preview

**Ground-truth**

Screenshot 2026-07-09 at 12 37 06


dolfim-ibm (approved):
nice!

PR #298, opened 2026-07-11
fix: resolve core-14 font aliases to their standard metrics (2)

PeterStaar-IBM:
fixed the rendering for the test-case!

Screenshot 2026-07-11 at 07 37 02

superseding: https://github.com/docling-project/docling-parse/pull/294


pablopupo:
[core14-alias-no-widths-extended.pdf](https://github.com/user-attachments/files/29928979/core14-alias-no-widths-extended.pdf)

Here it is. One page with thirteen lines, one per row of the alias table, arial, couriernew, times and timesnewroman in regular, bold, italic, and bolditalic. It's the same construction as before where every font with no /Widths and no FontDescriptor, and each line starts with the BaseFont it exercises so a render is easy to eyeball. I verified every line on this branch against the matching AFM, the widths and the per-glyph advances both match to the third decimal.

Note that on main before #294 the nine non-courier lines collapse to the flat 500 fallback, but the courier ones already read 7.2pt there since couriernew substring-matches courier and every Courier style is 600 units wide, so asserting per-glyph advances catches more than asserting the width changed. Also the BaseFont spellings are the exact alias keys on purpose, producer names like Arial-BoldMT would longest-match the bundled Microsoft AFMs instead of the alias rows.

maxmnemonic (dismissed):

dolfim-ibm (approved):

PR #299, opened 2026-07-14
fix: emit glyph marker instead of fabricated encoding text for symbolic-font cmap misses

wittjeff:
Fixes the text-extraction corruption reported in docling-project/docling#3802 (Arabic في coming out as `#`). Full root-cause analysis: https://github.com/docling-project/docling/issues/3802#issuecomment-4971598661

## Problem

When a character code is not covered by `/Encoding/Differences` or the font's `/ToUnicode` CMap, `get_correct_character()` falls through to `get_character_from_encoding()`, which resolves the code through the declared base encoding and then **StandardEncoding**. For symbolic fonts (e.g. subsetted Arabic fonts, whose codes are assigned sequentially from 0x21 and bear no relation to Latin encodings) this *fabricates* ASCII text: an Arabic ligature glyph subsetted at code `0x23` whose ToUnicode entry the generator omitted — a common generator bug for one-glyph-to-many-codepoints ligatures — comes out as a literal `#`.

Because the fabricated character looks like plausible text, downstream consumers cannot detect the corruption (docling's `rate_text_quality()` gate specifically watches for `GLYPH<...>` markers, which never appear on this path).

## Fix

The unmapped-code fallback call sites in `get_correct_character()` are routed through a new `resolve_unmapped_character()`: when an explicit `/ToUnicode` (or predefined CID) CMap exists but does not cover the code, **and** the code cannot meaningfully resolve through the standard Latin encodings — the font is symbolic (`/FontDescriptor /Flags` bit 3) or composite (Type0, whose codes are CIDs) — it emits the existing `GLYPH<code>` marker instead of fabricating text from the encoding tables.

- The guard applies only to the *encoding-table* fallbacks. Authoritative sources stay ahead of it and are untouched: `/Differences`, the cmaps, and the known base-font tables (`fm.has(c)` — e.g. TeX math fonts resolving `∑`).
- Non-symbolic simple fonts keep the current encoding fallback unchanged (a partial ToUnicode on a Latin font still resolves the rest through WinAnsi/Standard, where e.g. `0x23` genuinely is `#`).
- Fonts with no ToUnicode at all are unchanged.
- With default decoding (`keep_glyphs=False`) the marker is already rendered as a space, so consumers see whitespace instead of a wrong character; with `keep_glyphs=True` the marker is visible and can trigger OCR fallback logic.
- `/FontDescriptor /Flags` parsing is hoisted into `init_font_flags()` and reused by `build_embedded_font_blob()`, which previously parsed it inline.

## Tests

`tests/test_tounicode_fallback.py` builds the PDFs in memory (same pattern as `test_standard_font_widths.py`): a simple TrueType font, text bytes `0x21 0x22 0x23`, ToUnicode covering `0x21/0x22`:

| variant | before | after |
|---|---|---|
| symbolic, ToUnicode omits 0x23 | `#` (fabricated) | `GLYPH<35>` |
| symbolic, ToUnicode maps `<23> <0641064A>` | في | في (unchanged) |
| non-symbolic, ToUnicode omits 0x23 | `#` | `#` (unchanged) |

The reference regression suite (`test_reference_documents_from_filenames`) passes unchanged, and a real-world Arabic PDF with complete ToUnicode (UAE labor law, ADJD mirror) extracts identically before/after.

### DIFF
--- a/src/parse/pdf_resources/page_font.h
+++ b/src/parse/pdf_resources/page_font.h
@@ -84,6 +84,7 @@ namespace pdflib
   private:
 
     std::string get_correct_character(uint32_t c);
+    std::string resolve_unmapped_character(uint32_t c);
     std::string get_character_from_encoding(uint32_t c);
 
     void init_encoding();
@@ -124,6 +125,8 @@ namespace pdflib
     void init_cmap(pdf_timings& timings);
     void init_cmap_resource();
 
+    void init_font_flags();
+
     void init_differences();
 
     void init_charprocs();
@@ -150,6 +153,7 @@ namespace pdflib
     std::string        encoding_name;
     font_encoding_name encoding;
     bool               has_explicit_encoding; // true if encoding was found in PDF, false if defaulted
+    bool               is_symbolic = false;   // /FontDescriptor /Flags bit 3 (PDF 32000-1 table 123)
 
     font_subtype_name  subtype;
 
@@ -575,7 +579,7 @@ namespace pdflib
     else if(cmap_initialized and cmap_numb_to_char.count(c)>0)
       {
         return cmap_numb_to_char.at(c);
-      }         
+      }
     else if(bfonts.has_corresponding_font(font_name))
       {
         // check if the font-name is registered as a 'special' font, eg
@@ -591,7 +595,7 @@ namespace pdflib
         if(has_explicit_encoding &&
            (encoding == MACROMAN || encoding == MACEXPERT || encoding == WINANSI || encoding == STANDARD))
           {
-            return get_character_from_encoding(c);
+            return resolve_unmapped_character(c);
           }
         else if(fm.has(c))
           {
@@ -605,7 +609,7 @@ namespace pdflib
 	      << "; Encoding: " << to_string(_encoding)
 	      << "; font-name: " << font_name;
 	    */
-	    return get_character_from_encoding(c);
+	    return resolve_unmapped_character(c);
 	  }
 	else
 	  {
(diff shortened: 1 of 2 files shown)


trehansalil:
When will this be merged?

yalsaffar:
Noticed CI hasn't run on `0ec063c` yet, so I built this branch and `main` next to each other and diffed the output over 104 Arabic PDFs — 10 pages each, `iterate_cells(TextCellUnit.LINE)`, default decoding. Same container, same tree, only `page_font.h` swapped, and the binaries did come out different (`md5 3fd2f13bbc25` for main `40d51f3` vs `20c2cb5da1cd` for `0ec063c`), so it's a real A/B.

| | result |
|---|---|
| documents | 104 |
| byte-identical | 104 / 104 |
| pages compared | 997 |
| pages changed | 0 |

Nothing moved. I leaned on byte-identity rather than counting markers, since with `keep_glyphs=False` a marker comes out as a space — so a fabricated `#` turning into whitespace would still show up as a diff, and none did.

Worth saying the corpus isn't short of relevant fonts: 2,392 distinct ones, and 1,862 of those (77.8%) are symbolic or Type0 with a `/ToUnicode`. Mostly TrueType, 395 Type0, 93.5% subsetted.

Obvious caveat — this only tells you the change is safe, not that it works. The new branch never fired here, so it says nothing about the `في` → `#` fix itself. Your in-memory tests cover that much better anyway.

One thing that might be a fourth case, though, outside the three you list. With `set_loglevel("warning")` these files throw a lot of:

```
page_font.h:2306  WARN| could not find cmap in '/ToUnicode'
page_font.h:870   WARN| font-encoding not defined, falling back to STANDARD
```

Symbolic font, `/ToUnicode` there, but nothing usable parsed out of it — so `cmap_initialized` is false, the guard gets skipped, and it still fabricates. If that shape turns out to be common in the wild the fix might reach fewer documents than #3802 suggests. Happy to dig into it if that's worth knowing.

Separate thing, but same repo: 7.11.0 hit `MemoryError: std::bad_alloc` around page 11 on every document I tried, regardless of size — smells like memory piling up across `get_page()` rather than any particular page. Rebuilding the parser every 8 pages got me past it. Can file that properly with a reproducer if you want.

src/parse/pdf_resources/page_font.h
```diff
@@ -627,16 +631,57 @@ namespace pdflib
...
+    // can detect the unresolved glyph.
+    if(cmap_initialized and (is_symbolic or subtype==TYPE_0))
+      {
+        unknown_numbs[c] += 1;
+
+        LOG_S(WARNING) << "Symbol not in the cmap of a symbolic or composite font: "
+                       << int(c) << "; font-name: " << font_name
+                       << "; emitting glyph marker instead of encoding fallback";
+
+        return "GLYPH<" + std::to_string(c) + ">";
```
PeterStaar-IBM, line 679: this should only be returned if keep_glyphs in the config is true (see https://github.com/docling-project/docling-parse/blob/main/src/parse/config.h#L62)

PeterStaar-IBM (changes requested):

src/parse/pdf_resources/page_font.h
wittjeff: Same situation as #302: the marker does respect `keep_glyphs`, via the strip #231 added in `pdf_states/text.h` — with `keep_glyphs=false` (the default) any `GLYPH<...>` string from `page_font.h` is reduced to a space before it reaches output, exactly like the pre-existing markers in this file. `page_font` has no access to the decode config, so gating at emission would require plumbing it through the resource layer.

Added an explicit test (`test_default_config_strips_glyph_markers`) asserting that under the default config neither the marker nor a fabricated '#' reaches the extracted text. Also rebased onto main — full suite 152/152 green with `uv run pre-commit run --all-files` passing.

PeterStaar-IBM: wonderful!

PeterStaar-IBM (approved):
lgtm!

wittjeff:
@yalsaffar I am digging into your comments. It'd be great if you could share your Arabic testing docs corpus.

wittjeff:
Thank you — that's a genuinely rigorous A/B, and the byte-identity framing (a fabricated `#` collapsing to whitespace under `keep_glyphs=False` would still diff) is exactly the right way to prove it's non-regressive. Agreed on the caveat: it shows the change is safe, not that the guard fires on that corpus — those 104 all carry usable `/ToUnicode`, so they never reach the fabrication path. The in-memory tests are what exercise the `في` → marker behaviour.

**On your fourth case — you're right, and it's a real gap.** I traced it: the guard is

```cpp
if(cmap_initialized and (is_symbolic or subtype==TYPE_0))
    return "GLYPH<...>";
return get_character_from_encoding(c);   // fabricates
```

so it only protects fonts where a `/ToUnicode` cmap *parsed*. Your shape — symbolic font, `/ToUnicode` present but nothing usable extracted — sets `cmap_initialized = false`, skips the guard, and falls straight into `get_character_from_encoding`, which fabricates. So exactly as you say: those `could not find cmap in '/ToUnicode'` + `falling back to STANDARD` files are unprotected, and if that shape is common the fix reaches fewer docs than #3802 implies.

I'll open a follow-up that widens the guard to emit the marker for symbolic/Type0 fonts whenever no authoritative mapping resolved — i.e. drop the `cmap_initialized` requirement for that font class (STANDARD-encoding output is never meaningful for a symbolic font anyway), with a test built from one of those failed-`/ToUnicode` files. If you can point me at (or share the font descriptor from) one of the `page_font.h:2306` cases you hit, I'll fold it in as the fixture. Very happy to have you dig in alongside.

**On the `std::bad_alloc` around page 11 — that's a known one, already fixed.** We root-caused that exact class recently (docling#3829): `page_cells.h` reserves ~1M cells per page decode (≈7×352 MB), which accumulates across `get_page()` and throws `std::bad_alloc` — matching your "piles up rather than any particular page" and your rebuild-every-N-pages workaround precisely. It's fixed in **docling-parse 7.12.0 (PR #311)**; on 7.12.1 the whole-conversion peak commit dropped from ~3470 MB to ~1791 MB. You're on 7.11.0, which is before that — upgrading to 7.12.1 should clear it without the parser-rebuild workaround. If it still reproduces there, it's a distinct issue and a reproducer would be very welcome.

Issue #238, opened 2026-03-08
DoclingParseV4 backend produces glyph IDs (/gidXXXXX) for PDFs with custom embedded fonts without ToUnicode map


PeterStaar-IBM:
link to original issue: https://github.com/docling-project/docling/issues/3081

wittjeff:
I spent some time on this and want to share two findings: the attached file does **not** reproduce for me, but the underlying defect class is real and reproducible synthetically — and half of it is already fixed on `main`.

## The attached file extracts cleanly

`12-gestione-del-parto-pretermine-min-2.pdf` (115 pages) produces **zero** `/gidNNNNN` tokens under every combination I tried (the only "gid" substring in the output is the Italian word *egida*):

| docling-parse | driver | result |
|---|---|---|
| 5.11.0 (docling 2.76.0 pin) | docling `DoclingParseDocumentBackend` | clean |
| 5.11.0 (docling 2.76.0 pin) | docling `DoclingParseV4DocumentBackend` (deprecated shim) | clean |
| 4.4.0 | direct `DoclingPdfParser` | clean |
| current `main` | direct `DoclingPdfParser` | clean |

@a-n-d-r-e-a-b (reporter of docling#3081) — could you re-check this exact attachment against your environment, or attach a file that still reproduces? It's possible the failing corpus differs from the minimized attachment.

## The defect class is real

A minimal synthetic — a font whose `/Encoding/Differences` uses subset glyph-index names, `/Differences [ 65 /gid00043 /gid00049 /gid00041 ]`, drawn as `(ABC) Tj`, with a `/ToUnicode` CMap mapping the codes to `C`,`I`,`A`:

1. **Released 5.11.0 emits `/gid00043/gid00049/gid00041` even though the ToUnicode CMap is present and valid.** Current `main` extracts `CIA` — `init_differences()` now consults the CMap first per PDF 32000-1 §9.10.2 — so this half is already fixed and will ship with the next release.
2. **Without a ToUnicode CMap** (this issue's title case), current `main` still emits `gid00043gid00049gid00041` as ordinary reading text. Because that's plausible-looking text rather than a `GLYPH<...>` marker, downstream quality gates can't detect it and OCR fallback never triggers. I've opened #302 to emit a `GLYPH<name:gid00043>` marker in that case (same philosophy as #299).

[...]

PR #302, opened 2026-07-14
fix: emit glyph marker for glyph-index names instead of leaking them as text

wittjeff:
Addresses the glyph-name half of #238.

## Problem

Subset generators (FontForge, fontTools, mPDF, ...) name glyphs by bare index: `/gid00043`, `/g43`, `/glyph43`. Such a name identifies the glyph inside the embedded font program but carries no reading text. When `/Encoding/Differences` assigns these names and no `/ToUnicode` CMap covers the code, `init_differences()`'s last-resort fallback kept the raw name as reading text:

```
/Differences [ 65 /gid00043 /gid00049 /gid00041 ]  →  "gid00043gid00049gid00041"
```

Because that is plausible-looking text rather than a `GLYPH<...>` marker, downstream consumers cannot detect the corruption — docling's `rate_text_quality()` gate never fires and no OCR fallback triggers. This is the same fabricated-text class as #299, reached through the glyph-*name* path instead of the encoding path.

(Note: the with-ToUnicode variant of #238 is already fixed on `main` — `init_differences()` consults the CMap first per PDF 32000-1 §9.10.2. Released 5.11.0 still emits `/gidNNNNN` even when a valid ToUnicode exists. And the file attached to #238 does not reproduce for me on 5.11.0 or `main` — details in the issue thread.)

## Fix

The last-resort fallback (`resolve_unknown_name`) now detects pure glyph-index names — `(gid|glyph|g|cid|index)\d+`, case-insensitive, full-match — and emits a `GLYPH<name:gid00043>` marker instead of the name. Under the default `keep_glyphs=false` the marker is stripped to a space, so no fabricated text reaches consumers; with `keep_glyphs=true` the glyph identity is preserved for diagnostics.

Deliberately unchanged:

- a valid `/ToUnicode` CMap stays authoritative ahead of the guard (method 1 of §9.10.2);
- meaningful unknown names (custom ligatures like `/Th`, `/ft`) keep the existing name-as-text fallback;
- glyph-name based methods (AGL tables, `uniXXXX`, ligature join/decompose, `/C<digits>`) all run before the guard.

Recovering the *actual* text for these fonts would require reverse-looking-up the embedded font program's cmap table (gid → Unicode) — worth discussing as a follow-up; this PR makes the corruption detectable so OCR fallback can kick in meanwhile.

## Tests

`tests/test_gid_glyph_names.py` builds the PDFs in memory (same pattern as #299's tests): gid-named Differences with and without ToUnicode, name variants (`g43`, `glyph49`, `index41`), and a regression guard for ligature-style names. Full suite passes (88 tests).

### DIFF
--- a/src/parse/pdf_resources/page_font.h
+++ b/src/parse/pdf_resources/page_font.h
@@ -2215,16 +2215,33 @@ namespace pdflib
 	and cmap_numb_to_char.at(numb)!=replacement_char;
     };
 
+    // Subset generators (FontForge, fontTools, mPDF, ...) name glyphs by
+    // bare index: /gid00043, /g43, /glyph43, /cid43, /index43. Such a name
+    // identifies the glyph inside the embedded font program but carries no
+    // reading text, so keeping it fabricates plausible-looking garbage
+    // ('gid00043gid00049...') that downstream quality gates cannot detect
+    // (docling-project/docling-parse#238).
+    std::regex re_gid(R"((gid|glyph|g|cid|index)\d+)", std::regex::icase);
+
     // Last-resort for glyph-names that neither the /ToUnicode cmap nor
     // any glyph-table could resolve (eg custom ligatures like /Th, /ft
     // or /tt in a font without cmap): keep the glyph-name itself without
-    // the leading '/' and any '.suffix' as the reading text.
-    auto resolve_unknown_name = [](const std::string& name, const std::string& name_)
+    // the leading '/' and any '.suffix' as the reading text — unless the
+    // name is a pure glyph-index, which becomes a GLYPH marker so the
+    // unresolved glyph stays detectable.
+    auto resolve_unknown_name = [&re_gid](const std::string& name, const std::string& name_)
     {
       std::string result = name_;
       if(result.empty())
 	{
 	  result = (name.size()>0 and name[0]=='/')? name.substr(1) : name;
+	}
+
+      if(std::regex_match(result, re_gid))
+	{
+	  LOG_S(WARNING) << "glyph-index name " << name
+			 << ": emitting marker 'GLYPH<name:" << result << ">'";
+	  return "GLYPH<name:"+result+">";
 	}
 
       LOG_S(WARNING) << "unknown glyph-name " << name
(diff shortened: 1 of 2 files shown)


src/parse/pdf_resources/page_font.h
```diff
@@ -2218,18 +2218,35 @@ namespace pdflib
...
       if(result.empty())
 	{
 	  result = (name.size()>0 and name[0]=='/')? name.substr(1) : name;
 	}
 
+      if(std::regex_match(result, re_gid))
+	{
+	  LOG_S(WARNING) << "glyph-index name " << name
+			 << ": emitting marker 'GLYPH<name:" << result << ">'";
+	  return "GLYPH<name:"+result+">";
```
PeterStaar-IBM, line 2244: this should only be returned if `keep_glyphs` in the config is `true` (see https://github.com/docling-project/docling-parse/blob/main/src/parse/config.h#L62)

PeterStaar-IBM (changes requested):

PeterStaar-IBM:
@wittjeff Happy to see your PR's: please always run the linter `uv run pre-commit run --all-files`

src/parse/pdf_resources/page_font.h
wittjeff: It does respect `keep_glyphs` — via the mechanism #231 added: any `GLYPH<...>` string returned by `page_font.h` is stripped to a space in `pdf_states/text.h` (`if((not config.keep_glyphs) and chars_.rfind("GLYPH<", 0) == 0)`) when `keep_glyphs` is false, the same path the existing markers in this file use (`GLYPH<c=...,font=...>`, `GLYPH<UNKNOWN>`, `GLYPH<numb>`). `page_font` itself has no access to the decode config, so gating at emission would mean plumbing the config through the resource layer; it would also break `init_space_index()`, which probes `get_string(32) == " "` to locate the space glyph.

I've added an explicit test (`test_default_config_strips_gid_markers`) asserting that with the default config neither the marker nor the fabricated gid text reaches the output. Happy to plumb the config through instead if you'd prefer the gate at emission.

wittjeff:
Linter run as requested — `uv run pre-commit run --all-files` now passes all hooks (Ruff format/lint, MyPy, uv-lock). Also rebased onto current main (v7.12.1): the branch predated #300/#310 and the pinned groundtruth revision, which is why the earlier CI run was stale. Full suite is 153/153 locally after the rebase.

src/parse/pdf_resources/page_font.h
PeterStaar-IBM: ok, perfect, thanks for reminding me ;). Love the contributions!

PeterStaar-IBM (approved):
lgtm!

PR #305, opened 2026-08-04
feat: added the rendering-regression with pypdfium

PeterStaar-IBM:
## Fix page-size rounding and honor page rotation when rendering

  Renders of some pages came out slightly undersized, shifted, or in the wrong
  orientation. Both causes are in how the page box reaches the renderer.

  The visualization comparison is shown here: https://huggingface.co/datasets/docling-project/regression-dataset-for-docling-parse/tree/main/visualizations

  ### What was wrong

  1. **Page boxes were truncated to integers.** `size_instruction` stored
     `media_bbox`/`crop_bbox` as `std::array<int, 4>`, so a page box of
     `595.2756 x 841.8898` became `595 x 841`. That both shifts the page origin
     and shrinks the page by up to a point, which the renderer turns into a
     misplaced and undersized canvas.
  2. **Page boxes were not normalized.** A rectangle may be written with any pair
     of opposite corners (ISO 32000-1, 7.9.5). A page box written top-down has a
     negative extent, which propagated into every derived page size.
  3. **`/Rotate` was dropped.** The page angle (ISO 32000-1, Table 30) never left
     the parser, so a page marked 90 or 270 rendered sideways relative to how
     every viewer displays it.

  ### What changed

  - `size_instruction` now carries real-valued `media_bbox`/`crop_bbox` and the
    page's `angle`; `page_dimension` normalizes boxes to lower-left/upper-right.
  - `pdf_decoder<PAGE>` captures `/Rotate` **before** `rotate_contents()`
    normalizes it away — render instructions stay in unrotated page space, so the
    renderer is the one that needs the angle.
  - `resolve_canvas_size()` takes doubles and rounds canvas extents **up**
    (`pixels_for_extent`, with an epsilon to absorb subtraction noise). This
    matches PDFium, so our canvas dimensions line up with a pypdfium2 render of
    the same page.
  - The Blend2D renderer allocates its canvas in unrotated page space and rotates
    the finished canvas clockwise by whole quarter turns before handing it out.
    Quarter turns permute pixels exactly — no resampling, no precision loss. The
    rotation is one-shot, so `get_canvas()` followed by `save()` rotates once.
    Non-multiple-of-90 angles are logged and treated as unrotated.
  - `angle` is exposed through the pybind JSON as `size_instruction.angle`.

  ### Tests

  - New `tests/test_pypdfium_render.py`: compares our renders against pypdfium2.
    Marked `pypdfium` and non-failing — it reports rather than gates.
  - `tests/rendering_regression.py` gained resampling-based comparison so pages
    differing by a pixel or two in size are compared on top of each other instead
    of drifting apart toward the page edge; size mismatch is still reported
    separately. Three-panel comparison images are written per the new
    `--render-visualizations` option (`above-tolerance` by default).
  - `test_threaded_render.py` expectations moved from `round()` to `math.ceil()`
    to match the new canvas sizing.
  - `tests/README.md` documents the new test surface.

### DIFF
--- a/app/pybind_parse.cpp
+++ b/app/pybind_parse.cpp
@@ -74,6 +74,7 @@ namespace
       size["type"] = instruction_name(pdflib::SIZE_INSTRUCTION);
       size["media_bbox"] = instr.media_bbox;
       size["crop_bbox"] = instr.crop_bbox;
+      size["angle"] = instr.angle;
       root["size_instruction"] = size;
     }
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,7 +167,8 @@ python_version = "3.10"
 module = [
     "tabulate.*",
     "botocore.*",
-    "boto3.*"
+    "boto3.*",
+    "pypdfium2.*"
 ]
 ignore_missing_imports = true
 
--- a/src/parse/page_items/page_dimension.h
+++ b/src/parse/page_items/page_dimension.h
@@ -291,6 +291,20 @@ namespace pdflib
 	  }
       }
 
+    // A rectangle may be written with any pair of opposite corners, and shall
+    // be normalised so that the first corner is the lower-left one
+    // (ISO 32000-1, 7.9.5). Without this, a page box written top-down has a
+    // negative extent, which propagates into every page size derived from it.
+    if(result[0] > result[2])
+      {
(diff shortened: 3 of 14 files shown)


dolfim-ibm (approved):
lgtm

PR #306, opened 2026-08-05
 feat: render PDF shadings (sh), parse the full ExtGState, and add rendering quality benchmarks

PeterStaar-IBM:
## Summary

  Two rendering gaps and the reporting needed to see them.

  1. **The `sh` shading operator was a no-op.** It was recognised and logged, then dropped — there was no `/Shading`
  resource decoding, no PDF function evaluator, and no render instruction that could carry a gradient. Pages with
  shaded backgrounds rendered without them.
  2. **`/ExtGState` was only partially parsed.** Line parameters, constant alpha, blend mode and soft-mask state are
  now read, and alpha is honoured by the renderer.
  3. **A quality benchmark** comparing every page of the regression corpus against pypdfium2, with a normalized 0–1
  metric and per-page difference visualisations.

  ## `sh` shading support

  `stream.h` previously held:

  ```cpp
  case pdf_operator::sh:
    {
      LOG_S(INFO) << "executing " << to_string(name);
      //current_graphic_state().sh(parameters);   // never written
    }
  ```

  New files:

  | File | Purpose |
  | --- | --- |
  | `src/parse/pdf_resources/pdf_function.h` | PDF function evaluator (ISO 32000-1, 7.10): types 0 (sampled), 2 (exponential), 3 (stitching) and 4 (PostScript calculator, with a full RPN interpreter). Undecodable functions carry a reason string. |
  | `src/parse/pdf_resources/page_shading.h` | One `/Shading` entry — type, colour space, function(s), `/Coords`,
  `/Domain`, `/Extend`, `/BBox`. Axial and radial shadings resample into a 256-step colour ramp. |
  | `src/parse/pdf_resources/page_shadings.h` | The `/Shading` subdictionary, parent-linked like the other resource collections, so form XObjects and page-tree inheritance both resolve. |

  `shading_instruction` is the new render instruction; `renderer<BLEND2D>::render_shading` paints it as a
  `BLGradient`.

  ### Design decisions worth reviewing

  **Coordinates stay in shading space.** The instruction carries the CTM as a 6-element matrix and the renderer
  composes it with the page-to-canvas mapping into a Blend2D *style* transform. Transforming the axis endpoints
  instead would be wrong under a non-conformal CTM: an axial shading's colour is constant along the lines
  perpendicular to its axis *in shading space*, and an affine map does not preserve those angles.

  **`/BBox` becomes a clip path.** It bounds the paint exactly as a clip does, so it is transformed into page space
  and appended to the instruction's clip state, reusing the existing intersection machinery.

  **`/Extend false` uses a transparent guard stop.** Blend2D's pad mode is the only one that leaves the interior ramp

  **An inapplicable clip skips the paint.** The renderer supports only axis-aligned rectangular clips; when none of a
  shading's clip paths qualify it warns and paints nothing, because an unbounded `sh` would flood the page. Same
  reasoning gates the parse side on `config.keep_shapes`.

  **`sh` is handled by the stream decoder, not the graphics state**, since it needs the `/Shading` resources and the
  current clipping path — neither of which the graphics state owns. The never-defined `pdf_state<GRPH>::sh`
  declaration is removed.

  **`/Background` is deliberately unread** — 8.7.4.3 says it shall be ignored by `sh`.

  ### Diagnostics

  Every path that declines to paint now says why: an unresolvable resource name, an unsupported `/ShadingType` (named,
  e.g. `"Coons patch mesh (6)"`), an undecodable colour space, a function failure with the sub-function index, a
  degenerate axis, an invisible `/ca`, or a clip that cannot be applied. The previous behaviour was a silent omission.

  ### Not implemented

  - Shading types 1 and 4–7 (function-based and the mesh families) — reported, not painted.
  - Shading *patterns* (`/PatternType 2` through `scn`); this covers the `sh` operator only.
  - Sampled functions of more than one input (only reachable from type 1 shadings).

  ## Graphics-state parsing
  `/ExtGState` now reads `/LW /LC /LJ /ML /D /FL /CA /ca /BM /SMask /AIS`. Each entry carries a `has_*` presence flag,
  so a `gs` that omits a key leaves the current value alone instead of resetting it to a default — which is what the
  spec requires and what makes nested `q`/`gs`/`Q` behave.

  `blend_mode_name` and `soft_mask_state` are parsed but not composited. They are kept on the state so a document
  relying on them is identifiable in the logs rather than silently rendered as Normal. Constant alpha *is* honoured:
  `/CA` and `/ca` flow into the shape, bitmap and text instructions, and the Blend2D renderer skips paint below
  `1/512` alpha.

  New files:

  | File | Purpose |
  | --- | --- |
  | `src/parse/pdf_resources/pdf_function.h` | PDF function evaluator (ISO 32000-1, 7.10): types 0 (sampled), 2 (exponential), 3 (stitching) and 4 (PostScript calculator, with a full RPN interpreter). Undecodable functions carry a reason string. |
  | `src/parse/pdf_resources/page_shading.h` | One `/Shading` entry — type, colour space, function(s), `/Coords`, `/Domain`, `/Extend`, `/BBox`. Axial and radial shadings resample into a 256-step colour ramp. |
  | `src/parse/pdf_resources/page_shadings.h` | The `/Shading` subdictionary, parent-linked like the other resource collections, so form XObjects and page-tree inheritance both resolve. |

  `shading_instruction` is the new render instruction; `renderer<BLEND2D>::render_shading` paints it as a
  `BLGradient`.

  ### Design decisions worth reviewing

  **Coordinates stay in shading space.** The instruction carries the CTM as a 6-element matrix and the renderer
  composes it with the page-to-canvas mapping into a Blend2D *style* transform. Transforming the axis endpoints
  instead would be wrong under a non-conformal CTM: an axial shading's colour is constant along the lines
  perpendicular to its axis *in shading space*, and an affine map does not preserve those angles.

  **`/BBox` becomes a clip path.** It bounds the paint exactly as a clip does, so it is transformed into page space
  and appended to the instruction's clip state, reusing the existing intersection machinery.

  **`/Extend false` uses a transparent guard stop.** Blend2D's pad mode is the only one that leaves the interior ramp
  intact, so a non-extended end gets a zero-alpha stop that pad replicates outwards. The guard band is sized to stay
  sub-pixel.

  **An inapplicable clip skips the paint.** The renderer supports only axis-aligned rectangular clips; when none of a
  shading's clip paths qualify it warns and paints nothing, because an unbounded `sh` would flood the page. Same
  reasoning gates the parse side on `config.keep_shapes`.

  **`sh` is handled by the stream decoder, not the graphics state**, since it needs the `/Shading` resources and the
  current clipping path — neither of which the graphics state owns. The never-defined `pdf_state<GRPH>::sh`
  declaration is removed.

  **`/Background` is deliberately unread** — 8.7.4.3 says it shall be ignored by `sh`.

  ### Diagnostics

  Every path that declines to paint now says why: an unresolvable resource name, an unsupported `/ShadingType` (named,
  e.g. `"Coons patch mesh (6)"`), an undecodable colour space, a function failure with the sub-function index, a
  degenerate axis, an invisible `/ca`, or a clip that cannot be applied. The previous behaviour was a silent omission.

  ### Not implemented

  - Shading types 1 and 4–7 (function-based and the mesh families) — reported, not painted.
  - Shading *patterns* (`/PatternType 2` through `scn`); this covers the `sh` operator only.
  - Sampled functions of more than one input (only reachable from type 1 shadings).

  ## Graphics-state parsing

  `/ExtGState` now reads `/LW /LC /LJ /ML /D /FL /CA /ca /BM /SMask /AIS`. Each entry carries a `has_*` presence flag,
  so a `gs` that omits a key leaves the current value alone instead of resetting it to a default — which is what the
  spec requires and what makes nested `q`/`gs`/`Q` behave.

  `blend_mode_name` and `soft_mask_state` are parsed but not composited. They are kept on the state so a document
  relying on them is identifiable in the logs rather than silently rendered as Normal. Constant alpha *is* honoured:
  `/CA` and `/ca` flow into the shape, bitmap and text instructions, and the Blend2D renderer skips paint below
  `1/512` alpha.

  ## Quality benchmarks

  `docs/quality_benchmarks.md` reports the Blend2D renderer against pypdfium2 over 108 pages / 75 documents,
  reproducible with:

  ```console
  uv run pytest tests/test_pypdfium_render.py -q -s --render-visualizations all
  ```

  **A normalized `delta` in [0, 1].** Per pixel, the largest of the three channel differences divided by 255, averaged
  over the page. `1.0` is attainable (every pixel maximally inverted), which is what makes the number comparable
  across pages of different size. It replaces `mean_abs_error` in the visualisation filenames, so a plain alphabetical
  listing is also a worst-last ranking — the old unbounded score sorted `delta_16.308` between `delta_1.6` and
  `delta_2.0`.

  Two deliberate differences from `mean_abs_error`: alpha is excluded (both renders are flattened onto opaque white
  first, so their alpha channels are identical by construction and averaging over four channels only scales the result
  by ¾), and channels are combined with `max()` rather than a mean or luma weighting, which would report a
  pure-blue-instead-of-black glyph as 85 or 29 instead of 255.

  **The difference panel is no longer amplified.** It was brightened 8×, which made a near-identical page look as
  damaged as a broken one. It is now the raw absolute difference, so a grey level of 40 means the renders are 40 apart
  there, and two panels from different pages are directly comparable. The cost is that an accurate page shows an
  almost black panel — the honest depiction of two renders agreeing.

  **Tables are sorted worst-first** by `mean_abs_error`, in both the pytest output and the doc.

  Current state: median `delta` 0.0129, 94 % of pages below 0.05, with a thin tail of four pages. Two of those are
  characterised in the doc as known limitations rather than regressions — `device_n_black.pdf` (un-evaluated
  `/Separation` and `/DeviceN` tint transforms) and the form pages (the light-blue widget-bounds overlay that PDFium
  has no counterpart for).

  ## Testing

  `tests/test_shading.py` — 12 tests over PDFs built in memory, asserting on rendered pixels:

  - painting under `W` and `W*`, and no leakage after `Q`
  - a flipped `1 0 0 -1 0 h cm` and a scaled `cm`, verifying the axis follows the CTM and not just the clip
  - radial shadings and `/Extend [false false]`
  - `/Shading` inherited from a `/Pages` ancestor
  - an unclipped `sh` covering the whole page
  - four failure modes — unsupported shading type, unsupported function type, missing `/Coords`, missing resource —
  each of which must leave the page intact

  `tests/constants.py` bumps the regression dataset revision.


dolfim-ibm (approved):
lgtm

PR #310, opened 2026-08-07
feat: improving performance and robustness

PeterStaar-IBM:
## Summary

  This PR improves parsing throughput and memory behavior by reducing expensive font-resource JSON materialization, reusing already sanitized cells for
  line-cell generation when the decode settings are equivalent, and avoiding redundant cell processing work. It also adds finer-grained timing keys for page
  decoding, font decoding, sanitization, and word/line cell creation so performance analysis can identify the remaining hot spots more directly.

  The build now forces Blend2D release assertion behavior for non-Debug artifacts and adds wheel checks to prevent shipping Blend2D debug assertion code.
  The performance tooling was updated with better timing breakdowns, resumable comparison runs, PDF selection, parse/render content options, and refreshed
  Apple M3 Max benchmark reports.

  ## Changes

  - Read font metadata directly from QPDFObjectHandle instead of eagerly converting font dictionaries to JSON.
  - Add reusable QPDF object access helpers.
  - Reuse sanitized cells for line-cell output when the requested line-cell config matches the sanitized representation.
  - Split parser timings into more detailed nested buckets.
  - Avoid extra work when extracting one-page PDFs from one-page documents.
  - Add Blend2D release-mode compile/link settings and CI wheel assertion checks.
  - Improve benchmark scripts with config controls, reusable comparison CSV data, PDF filtering, and clearer timing breakdowns.
  - Add/update performance benchmark documentation and plots.


dolfim-ibm (dismissed):
lgtm

maxmnemonic (approved):
lgtm!

PR #315, opened 2026-08-11
fix: make Blend2D font fallback portable across distros

nassarofficial:
## Summary
- Broaden Blend2D system/fallback font discovery beyond Debian-only paths so non-embedded PDF fonts render on RHEL/Fedora and other distro layouts (well-known paths + bounded directory scan).
- Support `DOCLING_PARSE_FALLBACK_FONT` / `DOCLING_PARSE_CJK_FALLBACK_FONT` overrides, and optionally pick up fonts from `pdf_resources/fonts/fallback/` if present.
- Detect CJK font requests (non-ASCII names or known family tokens) and prefer a CJK-capable system face before the Latin fallback (avoids `.notdef` boxes when Noto CJK etc. are installed).

### DIFF
--- a/src/render/blend2d_font_resolver.h
+++ b/src/render/blend2d_font_resolver.h
@@ -10,6 +10,8 @@
 #endif
 #include <loguru.hpp>
 
+#include <resources.h>
+
 #include <algorithm>
 #include <array>
 #include <cctype>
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM (approved):
lgtm!

PR #320, opened 2026-08-13
feat(render): add tiling patterns, clipping, Type3 glyphs and CID text

nassarofficial:
Rebuilt on top of the shading and colour work already in `main`, keeping
the render paths that work does not cover. A `/Pattern` fill, a curved
clip, a packed palette image, a Type3 page, and a CID font with no `/DW`
were each dropping or distorting content.
### Tiling patterns
- Replay the pattern cell on a lattice derived from the page box through
  the inverse pattern transform, so it extends in both directions
- Anchor pattern space to the page's default space, not the CTM at the fill
- Skip an unresolved `/Pattern` fill instead of painting placeholder black,
  without skipping clip capture / path reset
### Non-rectangular clipping
- Rasterise curved clip paths into an A8 coverage mask (Blend2D only clips
  rectangles; a crescent previously became its bounding box)
- Apply the mask with an explicit per-pixel multiply (`DST_IN` skips the
  pixels that need clearing)
### Indexed images and soft masks
- Unpack sub-byte `/Indexed` samples before palette lookup (packed 4-bit
  bytes were treated as one index each)
- Pass palette indices through untouched instead of scaling them as
  continuous tone
- Unpack sub-byte soft masks (1 bpc stencils were rejected)
### Type3, CID widths, text render modes
- Rasterise Type3 `/CharProcs` (image-mask and filled paths) onto the
  glyph quad instead of decoding and discarding them
- Default CID `/DW` to 1000 per spec, and only on composite fonts (applying
  it to every font short-circuited base-14 metric lookup)
- Honour stroke text render modes 1, 2, 5 and 6 so synthesised bold is drawn

### DIFF
--- a/src/parse.h
+++ b/src/parse.h
@@ -106,6 +106,8 @@
 
 #include <parse/pdf_resources/page_shading.h>
 #include <parse/pdf_resources/page_shadings.h>
+#include <parse/pdf_resources/page_pattern.h>
+#include <parse/pdf_resources/page_patterns.h>
 
 #include <parse/pdf_resources/page_xobject_image.h>
 #include <parse/pdf_resources/page_xobject_form.h>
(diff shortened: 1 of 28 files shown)


PeterStaar-IBM (approved):
lgtm!

PR #321, opened 2026-08-14
feat(render): clip paths, masks, CMYK/Indexed images, font fallback, Symbol brackets

nassarofficial:
Follow-up render fixes for Linux.

### Script-aware font fallback and soft-mask grid resolution
When the name-resolved face cannot draw a run (e.g. Arabic through a Latin fallback), probe faces by script before giving up. Soft masks that are finer than their image (common for text-as-mask) now resolve onto the finer grid instead of downsampling the mask and smearing detail.

### Clip paths and text condensation
Subpaths captured by a single `W`/`W*` clip now compose into one path per clip operation; only separate clip operations intersect (ISO 32000-1 §8.5.4). Fixes multi-subpath clips (outline stencils, Qt “shape as clip, fill bbox”) that previously produced empty regions. Anisotropically scaled text (`Tz`, non-uniform text matrix) is fitted to the cell width in both Blend2D and FreeType paths.

### Stencil masks (`/Mask`) with CCITT decoding
`/Mask` streams (1-bit stencils that knock pixels out, inverse of `/SMask`) are now composited through the same resample path as soft masks. CCITTFax-compressed stencils are decoded via the vendored CCITT decoder, honouring `/K` and `/BlackIs1`.

### Indexed palettes over tint bases
`/Indexed` colour spaces built on `/Separation` or `/DeviceN` now convert palette entries through the base tint transform at load time (e.g. QR codes as Indexed-over-Separation no longer render as placeholders).

### CMYK polarity
Default `/DeviceCMYK` `/Decode` is the identity (ISO table 89). CMYK inversion applies only when the Adobe-inverted convention is positively established — fixes CMYK JPEG2000 negatives and accidental double-inversion elsewhere.
### Symbol bracket pieces and `.notdef` guards
Legacy Symbol bracket/brace/parenthesis piece codes (Adobe GL private-use area) are rewritten to Unicode U+239B..U+23AD at glyph-list load time. Glyph-identity recovery that maps every code to glyph 0 is rejected; FreeType skips drawing `.notdef` outlines while preserving advance.

### DIFF
--- a/app/pybind_parse.cpp
+++ b/app/pybind_parse.cpp
@@ -138,6 +138,7 @@ namespace
       pybind11::dict row;
       row["type"] = instruction_name(pdflib::BITMAP_RENDER_INSTRUCTION);
       row["xobject_key"] = instr.get_key();
+      row["source"] = pdflib::to_string(instr.get_source());
       row["shape"] = instr.get_shape();
       row["pixel_format"] = pixel_format_name(instr.get_pixel_format());
       row["image_mask"] = instr.is_image_mask();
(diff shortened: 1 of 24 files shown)


PeterStaar-IBM (approved):
lgtm!

PR #274, opened 2026-05-20
feat: release unused python memory

PeterStaar-IBM:
## Summary

  This change introduces a configurable native memory trimming policy to reduce allocator fragmentation during page processing.
  
  resolving: https://github.com/docling-project/docling-parse/issues/273, https://github.com/docling-project/docling-parse/issues/227

```
uv run python ./perf/run_scaling.py --threads "1,4,8,12" docling-project/performance-dataset-bo767

backend           threads      wall_time (s)  vs threaded(1)    vs pypdfium2 (1t)      pages/sec    ms/page
----------------  ---------  ---------------  ----------------  -------------------  -----------  ---------
pypdfium2 (1t)    -                 1043.72   0.30x             1.00x                       52.3      19.12
docling threaded  1                  308.543  1.00x             3.38x                      176.9       5.65
docling threaded  4                   85.432  3.61x             12.22x                     638.9       1.57
docling threaded  8                   47.727  6.46x             21.87x                    1143.7       0.87
docling threaded  12                  37.659  8.19x             27.72x                    1449.4       0.69
```

```
uv run python ./perf/run_scaling.py --threads "1,4,8,12" docling-project/performance-dataset-cornercases

backend           threads      wall_time (s)  vs threaded(1)    vs pypdfium2 (1t)      pages/sec    ms/page
----------------  ---------  ---------------  ----------------  -------------------  -----------  ---------
pypdfium2 (1t)    -                  459.439  0.34x             1.00x                      245.8       4.07
docling threaded  1                  155.231  1.00x             2.96x                      727.4       1.37
docling threaded  4                   67.277  2.31x             6.83x                     1678.4       0.6
docling threaded  8                   63.39   2.45x             7.25x                     1781.3       0.56
docling threaded  12                  64.138  2.42x             7.16x                     1760.5       0.57
```

PR #326, opened 2026-08-21
fix: open input PDFs with UTF-8 path semantics on Windows

### LINKED ISSUE #324: [Bug] docling-parse cannot open PDFs whose path contains non-ASCII (Unicode) characters on Windows — affects ALL such files (regression)
## Environment
- OS: Windows 10 (10.0.19045), Simplified-Chinese locale (GBK ANSI codepage 936)
- Python: 3.12.9 (conda env)
- docling: 2.120.3
- docling-parse: 7.14.0
- docling-core: 2.91.0
- Install path: `conda upgrade --all -y` + `uv pip install -U docling rapidocr onnxruntime`

## Summary
The native C++ `docling-parse` backend fails to **open** any PDF whose **path or filename contains non-ASCII characters** (e.g. Chinese `导出的条目`, `等`). Every such file immediately fails with `docling-parse could not load document` in ~0.03s, before any conversion model runs. The same file placed at an ASCII-only path converts fine.

This is a blocker: on a Chinese-locale Windows machine it makes conversion impossible for any file under a non-ASCII directory (a commonplace situation, e.g. `X:\导出的条目\...`).

## Steps to reproduce
1. Take any valid PDF and place it at a path containing Chinese/non-ASCII characters:
   `X:\导出的条目\导出的条目\tmp_work\2005 - Chiou 等 - ...pdf`
2. Load it with `DoclingPdfParser`:
   ```python
   from docling_parse.pdf_parser import DoclingPdfParser
   parser = DoclingPdfParser(loglevel="fatal")
   key = "key=<path>"
   ok = parser.parser.load_document(key=key, filename=str(path).encode("utf8"),
                                    password=None, keep_qpdf_warnings=False)
   print(ok)  # -> False
   ```
3. Copy the **same file** to an ASCII-only path `C:\Windows\Temp\ascii_test_docling.pdf` and repeat:
   ```python
   ok = parser.parser.load_document(key=key2, filename=b"C:\\Windows\\Temp\\ascii_test_docling.pdf", ...)
   print(ok)          # -> True
   parser.parser.number_of_pages(key=key2)  # -> 3
   ```

Native log line at failure:
```
document.h:196    ERR| could not open file: X:\导出的条目\导出的条目\tmp_work\2005 - Chiou 等 - ...pdf
docling_parser.h:181    ERR| could not process document for key=key=X:\导出的条目\导出的条目\tmp_work\...pdf
```

## CLI-level impact
```
docling "X:\导出的条目\导出的条目\2025 - Pan 等 - Flexible and.pdf" \
  --to md --output <dir> --ocr --image-export-mode referenced --enrich-formula ...
WARNING docling.document_converter: Input document ... is not valid.
WARNING docling.cli.main: ... failed to convert.
  [Failure Detail] ... docling-parse could not load document <sha>:
  Failed to load document with key key=X:\导出的条目\导出的条目\....
```
A follow-up `PermissionError [WinError 32]` during temp-dir cleanup also surfaces because the file handle opened by the native parser is not released after the failed load.

## Root-cause hypothesis
The native parser opens the PDF with a **narrow byte path (UTF-8-encoded)**. On Windows, filesystem APIs called with a narrow string interpret it using the **ANSI codepage** (GBK/936 for Simplified Chinese), not UTF-8. Any non-ASCII byte sequence therefore fails to resolve → `could not open file`. The fix is to open files via the Windows **wide-char (UTF-16) API** (or use `GetFullPathNameW`/`std::filesystem` with the appropriate conversion) so arbitrary Unicode paths work.

## Regression note
The author has previously converted ~10,000 PDFs with docling with only a handful of errors, using the same directory tree. After updating via `conda upgrade --all -y` + `uv pip install -U docling ...`, **all** files including previously-successful ones now fail with this same "could not open file" error. That strongly suggests the non-ASCII path handling regressed in a recent `docling-parse` release.

## Suggested fix / workaround
- Workaround for users today: copy each PDF to an ASCII-only temp path, convert, then move the outputs back. (Verified working.)
- Proper fix: in `docling-parse` native code, open input files with Unicode-aware path handling on Windows (`_wfopen` / `std::ifstream` with wide path, or convert the UTF-8 path to UTF-16 before opening).

wittjeff:
Fixes #324.

### Bug

On Windows, any PDF whose path contains non-ASCII characters fails to load with `could not open file` — every file under e.g. a Chinese-named directory is unconvertible. Regression introduced in #274 (first shipped in v6.1.0): the previous `qpdf_document.processFile(filename.c_str())` treats narrow filenames as UTF-8 on Windows and converts to wide internally, while the `std::ifstream(std::string)` that replaced it hands the UTF-8 bytes to the ANSI codepage (GBK, cp1252, cp437, …), where they resolve to a nonexistent name.

Not locale-specific: reproduced on a cp437 (US English) host with docling-parse 7.15.0 — ASCII sibling path loads, CJK path fails, BytesIO route loads. Details and repro output in [#324 (comment)](https://github.com/docling-project/docling-parse/issues/324#issuecomment-5364501962).

### Fix

In `process_document_from_file`, build a `std::filesystem::path` from the UTF-8 bytes (as `char8_t`, C++20) under `_WIN32` so the open goes through the wide-char API. POSIX keeps byte-passthrough semantics unchanged (paths that are raw bytes rather than valid UTF-8 behave exactly as before).

### Verification

- **Windows, end-to-end**: built a cp312/win_amd64 wheel from this branch with the repo's own MinGW wheel recipe and installed it on the same cp437 host that reproduces the bug with stock 7.15.0. Result: `ascii path: OK / cjk path: OK (1 pages) / cjk via BytesIO: OK` — where stock fails the cjk path.
- **Linux**: full test suite passes; the two new tests in `tests/test_unicode_paths.py` (CJK path load, BytesIO path-independence) pass.

Caveat for reviewers: CI runs tests on Linux only, so the new regression test guards intent on POSIX and will only actually bite on Windows if a Windows test lane is added — the wheel-based verification above is the real Windows evidence.

### DIFF
--- a/src/parse/pdf_decoders/document.h
+++ b/src/parse/pdf_decoders/document.h
@@ -3,6 +3,7 @@
 #ifndef PDF_DOCUMENT_DECODER_H
 #define PDF_DOCUMENT_DECODER_H
 
+#include <filesystem>
 #include <fstream>
 #include <mutex>
 #include <optional>
@@ -189,8 +190,21 @@ namespace pdflib
 
     utils::timer timer;
 
-    // Read the file into a buffer
-    std::ifstream ifs(filename, std::ios::binary | std::ios::ate);
+    // Read the file into a buffer.
+    //
+    // The filename arrives UTF-8 encoded from the Python layer. On Windows a
+    // narrow ifstream hands it to the ANSI codepage (GBK, cp1252, ...), so any
+    // non-ASCII path fails to open (docling-parse#324; regression from #274,
+    // which replaced qpdf's UTF-8-aware processFile with this read). Build a
+    // std::filesystem::path from the bytes as UTF-8 so MSVC opens via the
+    // wide-char API; POSIX keeps byte-passthrough semantics either way.
+#ifdef _WIN32
+    std::filesystem::path fs_path(
+      std::u8string(filename.begin(), filename.end()));
+#else
+    std::filesystem::path fs_path(filename);
+#endif
+    std::ifstream ifs(fs_path, std::ios::binary | std::ios::ate);
     if(!ifs.is_open())
       {
         LOG_S(ERROR) << "could not open file: " << filename;
--- a/tests/test_unicode_paths.py
+++ b/tests/test_unicode_paths.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+"""Loading PDFs from paths that contain non-ASCII characters.
+
+The filename travels UTF-8 encoded from Python into the native parser, which
+must open it with UTF-8 semantics on every platform. On Windows a narrow
+byte-path open goes through the ANSI codepage instead, so any path with
+characters outside it fails to load (docling-parse#324) — on POSIX the bytes
+pass through and the bug is invisible, so these tests guard intent everywhere
+but only bite on Windows runners.
+"""
+
+from io import BytesIO
+
+from docling_parse.pdf_parser import DoclingPdfParser
+from tests.pdf_builder import build_pdf
+
+CONTENT = "BT /F1 12 Tf 72 720 Td (Title Case Workers) Tj ET"
+
+
+def _tiny_pdf() -> bytes:
+    return build_pdf(
+        [
+            "<< /Type /Catalog /Pages 2 0 R >>",
+            "<< /Type /Pages /Kids [3 0 R] /Count 1 >>",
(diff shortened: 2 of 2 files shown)


PeterStaar-IBM (approved):
lgtm!

PR #329, opened 2026-08-25
ci: isolate regression tests in their own job with a cached HF corpus

cau-git:
The HF regression corpus is a large download and does not need to be pulled for each Python in the matrix. Move the regression tests to their own workflow pinned to a single Python version, cache tests/data keyed on tests/constants.py so HF_DATASET_REVISION bumps invalidate it, and have the per-Python matrix ignore tests/test_regression_*.py.


PeterStaar-IBM (dismissed):
lgtm!

PeterStaar-IBM (approved):
wonderful!

PR #337, opened 2026-09-02
feat: read the tagged-PDF structure tree and link text cells to marked content

wittjeff:
Phase 1 of the tagged-first proposal for docling (docling-project/docling#4143 covers forms; the wider outline is linked from docling-project/docling#3988). Docling can only treat authored PDF semantics as authoritative if the parser exposes them, and until now docling-parse read `/AcroForm`, `/Outlines`, `/Lang` and `/ActualText` but not the logical structure tree (ISO 32000-2, 14.7).

**What this adds**

- **Structure tree.** `PdfDocument.get_structure()` returns a `PdfStructure`: the `/MarkInfo /Marked` flag, the `/RoleMap`, PDF 2.0 `/Namespaces`, and the root elements. Each `PdfStructureElement` carries the raw `/S` type (with `resolved_type(role_map)` to follow the role map), `/NS`, `/T`, `/Lang`, `/Alt`, `/ActualText`, `/E`, `/ID`, the 0-based page of `/Pg`, attribute objects keyed by owner (`/Layout`, `/List`, `/Table`, `/PrintField`, `/Artifact`, `/ARIA-1.1`, …), `/Ref` targets by object identity, a depth-first `order` (the logical content order, 14.8.2.5), and its kids in document order: marked-content references (`page`, `mcid`), object references (`/OBJR` to annotations, with the annotation subtype), or nested elements. Cycles and depth are bounded.
- **Marked-content linkage on cells.** Every glyph cell now records the `/MCID` of the innermost marked-content sequence it was drawn in and the `/Type` of an enclosing `/Artifact` sequence. The existing BDC/EMC stack that already served `/ActualText` does the bookkeeping; tags are stamped at EMC before ActualText substitution so surviving cells keep them. `PdfDocument.get_page_marked_content(page_no)` returns the tags aligned with `char_cells` by `index`, so no docling-core model change is needed to consume them. The C++ `PdfCell` also exposes `mcid` and `artifact_type`, and the JSON cell record gains two trailing fields (25, 26), read back when present.

**What it does not do yet**

- Named property lists (`/P /Name BDC` resolved through `/Properties`) are not inspected, matching the current `/ActualText` behaviour.
- `/ParentTree` and `/StructParents` are not used; kids are resolved from the tree down, which is sufficient for reading. Annotation `/StructParent` values are not exposed.
- Nothing is synthesized for an `/ActualText` span that draws no glyphs.

**Tests**

`tests/test_unit_structure_tree.py` builds a one-page tagged PDF in memory with `tests/pdf_builder.py`: a role-mapped heading with `/Lang`, a paragraph with a nested `/Span` carrying `/ActualText`, a figure with `/Alt` and a Layout attribute, a link element referencing its annotation through `/OBJR` and the paragraph through `/Ref`, and a pagination artifact. It checks the tree shape, attributes, kid resolution, order, the per-cell MCID and artifact tagging (including the substituted span keeping the inner MCID), and that an untagged file yields no structure and no tags.

Companion work: a docling stage that turns these elements into pre-labeled layout clusters is the next step and will reference this PR.


wittjeff:
Design context for this PR is now docling-project/docling#4146 (the tagged-first proposal beyond forms); this reader is its phase 1.

wittjeff:
Consumer of this reader: docling-project/docling#4148 (draft) derives layout, heading levels and alt text from the tree. The last push here added text and geometry to the marked-content tags and exposed them on threaded page results, which that stage needs because docling does not materialize char cells by default.

PR #342, opened 2026-09-04
fix: expose bookmark target page indices

### LINKED ISSUE #340: Expose target page indices in table-of-contents bookmarks
## Summary

  The table-of-contents APIs exposed by `docling-parse` do not preserve the target page for each bookmark. Consumers therefore cannot constrain bookmark resolution to headings on the intended page.

  ## Current behavior

  Both `docling_parser` and `docling_threaded_parser` return ToC entries with title/level information but without the bookmark destination page index.

  Downstream code must either:

  - match each bookmark against heading candidates from every page, which is unnecessarily expensive; or
  - use a separate PDF library to read outline destinations.

  The latter is undesirable because it expands the PDFium/pypdfium integration surface in downstream backends.

  ## Expected behavior

  Extend the native ToC result from both parser backends so every bookmark can carry its zero-based target page index when one is available.

  The API should preserve existing behavior for bookmarks whose destinations  cannot be resolved.

  ## Why this matters

  Page-aware bookmark matching:

  - prevents matching a bookmark to a similarly named heading on an unrelated page;
  - reduces candidate matching from document-wide to page-local;
  - lets Docling consume native `docling-parse` ToC data without separately
    querying PDFium.

  ## Scope

  - Update both `docling_parser` and `docling_threaded_parser`.
  - Keep their ToC schemas/semantics aligned.
  - Add regression tests for direct page destinations and nested outlines.
  - Consider resolving named destinations where supported.

  ## Related context

  - https://github.com/docling-project/docling/pull/4169
  - https://github.com/docling-project/docling/pull/4169#issuecomment-5538190417

ItsRoy69:
## Summary

  Expose zero-based target page indices for PDF table-of-contents bookmarks.

  This allows consumers to constrain bookmark matching to the intended page and avoids unnecessary document-wide fuzzy
  matching.

  ## Changes

  - Resolve direct `/Dest` page destinations.
  - Resolve `/GoTo` action destinations.
  - Preserve page-less behavior for named or unresolved destinations.
  - Expose the ToC accessor on threaded parser and renderer bindings.
  - Add regression coverage for top-level and nested bookmark page indices.

  ## Validation

  - Ruff check passed.
  - Ruff formatting passed.
  - Focused unit tests passed: `6 passed`.
  - `git diff --check` passed.

  ## Issue

  Closes #340

### DIFF
--- a/app/pybind_parse.cpp
+++ b/app/pybind_parse.cpp
@@ -1089,6 +1089,12 @@ PYBIND11_MODULE(pdf_parsers, m) {
 
     Returns:
         int: Number of pages in the loaded document.)")
+    .def("get_table_of_contents",
+         [](docling::docling_threaded_parser& self, const std::string& key) {
+           return self.get_table_of_contents(key);
+         },
+         pybind11::arg("key"),
+         "Retrieve the document outline, including zero-based target page indices.")
     .def("scheduled_number_of_pages",
          [](docling::docling_threaded_parser& self, const std::string& key) -> int {
            return self.scheduled_number_of_pages(key);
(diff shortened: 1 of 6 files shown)


ItsRoy69:
Its already fixed by #341

PR #348, opened 2026-09-11
fix: answer shape geometry queries in the frame of the text cells

### LINKED ISSUE #343: get_shape_lines() and get_connected_shape_bounding_boxes() return raw PDF coordinates while text cells are relative to the page boundary

wittjeff:
## Summary

`PageParseResult.get_shape_lines()`, `get_connected_shape_bounding_boxes()` and `intersects_with()` (docling's `has_content_in()`) answered in raw PDF user space, while the text cells, shapes and images of the same page are rotated by `/Rotate` and relative to the page boundary (the crop box by default), and `page_width` / `page_height` are the boundary size. On a page whose boundary does not start at the origin the queries were off by exactly the boundary origin; on a rotated page they were off by the rotation as well.

Fixes #343. Complementary to #344, which translates the two list-returning wrappers in Python: this moves the fix into the decoder so that `intersects_with()` and rotated pages are covered too. Happy to fold it into #344 if @bharm16 prefers to carry it.

## Cause

`pdf_decoder<PAGE>::get_shape_lines`, `get_connected_shape_bounding_boxes` and the `shapes` branch of `intersects_with` (`src/parse/pdf_decoders/page.h`) iterate `instructions.get_shape_instructions()`. The render instructions are deliberately kept in unrotated user space so the renderer can orient its own canvas (see `decode_dimensions()`), whereas `page_cells`, `page_shapes` and `page_images` go through `rotate_contents()` and `page_item_sanitator<PAGE_DIMENSION>`. The `chars` and `bitmaps` branches of `intersects_with` read the sanitized items and were already correct.

## Change

- `rotate_contents()` records the angle and delta it applied; `decode_page()` records the boundary origin the dimension sanitator subtracted, read after rotation so it is in the rotated space.
- `get_shape_lines()` classifies a segment as horizontal or vertical in the page frame, so a quarter-turn page reports its lines by their orientation on the page rather than in the content stream.
- `to_page_frame(bbox)` applies `utils::values::transform_bottomleft_bbox_inplace` with that angle and delta, then the boundary translation, to a raw-space box. The three queries pass each box through it. The instructions are untouched, so rendering is unaffected; pages with `/Rotate 0` and a boundary at the origin return exactly what they did before.
- The docstrings of the Python wrappers now state the frame.

## Test

`tests/test_unit_shape_queries_page_frame.py` builds the page from #344 in memory with `tests/pdf_builder.py` (media box 400x300 at the origin, crop box 300x200 at (50, 40), the word `Hello`, a filled underline under it, a stroked line 6 points lower and a small RGB image to the right of the word), parametrized over `/Rotate` 0/90/180/270. For each rotation the expected frame is computed in the test from the raw user-space geometry, and the text cell, the sanitized shapes, the bitmap and the three queries must all agree with it. `intersects_with()` is probed with `shapes=True` and with `bitmaps=True` at the expected boxes (must hit) and at the raw user-space location (must miss). A second test keeps the readable "underline sits under its word" assertions from #344 for the unrotated page.

Observed on 7.19.0 before the change, for the underline:

| | /Rotate 0 | /Rotate 90 |
|---|---|---|
| `page.shapes` (sanitized) | l 50, b 76, r 90, t 77 | l 76, b 210, r 77, t 250 |
| `get_connected_shape_bounding_boxes()` | l 100, b 116, r 140, t 117 | l 100, b 116, r 140, t 117 |
| `intersects_with(shapes=True)` at the sanitized box | False | False |

## Validation

Built the extension locally from this branch and ran:

- `tests/test_unit_shape_queries_page_frame.py`: 5 passed (four rotations, text + shapes + bitmap, plus the readable check). Against the 7.19.0 build the rotated and shifted cases fail as in the table above.
- The two existing shape-query tests in `tests/test_regression_threaded_parse.py` (`get_shape_lines_returns_only_visible_straight_lines`, `get_connected_shape_bounding_boxes`): 2 passed, run standalone because the regression module's Hugging Face download is not reachable from my network. The full regression suite runs in CI.

## Not changed

The shape branch of the dimension sanitator sets `is_contained_in_boundary=true` in its else branch, so it never filters shapes outside the boundary. Left alone here since fixing it changes regression output.

https://claude.ai/code/session_01SeLRiBBA9VkM81fUwUQfyt

### DIFF
--- a/docling_parse/pdf_parser.py
+++ b/docling_parse/pdf_parser.py
@@ -1269,8 +1269,10 @@ class PageParseResult:
     ) -> bool:
         """Return whether visible page content intersects bbox.
 
-        bbox may use top-left or bottom-left coordinates. Native code expects
-        page coordinates with bottom-left origin as [left, bottom, right, top].
+        bbox may use top-left or bottom-left coordinates, in the frame of the
+        text cells: relative to the page boundary and rotated by /Rotate.
+        Native code expects page coordinates with bottom-left origin as
+        [left, bottom, right, top].
         """
         bbox_bl = bbox.to_bottom_left_origin(page_height=self.page_height)
         left = min(bbox_bl.l, bbox_bl.r)
@@ -1292,7 +1294,11 @@ class PageParseResult:
         vertical: bool = True,
         tolerance: float = 1e-3,
     ) -> List[BoundingBox]:
-        """Return visible horizontal and/or vertical stroked shape segments."""
+        """Return visible horizontal and/or vertical stroked shape segments.
+
+        Boxes are in the frame of the text cells: bottom-left origin, relative
+        to the page boundary and rotated by /Rotate.
+        """
         return [
             _to_bounding_box(tuple(bbox))
             for bbox in self._require_page_decoder().get_shape_lines(
@@ -1307,7 +1313,11 @@ class PageParseResult:
         *,
         tolerance: float = 0.0,
     ) -> List[BoundingBox]:
-        """Return bboxes of visible shapes connected by overlapping bboxes."""
+        """Return bboxes of visible shapes connected by overlapping bboxes.
+
+        Boxes are in the frame of the text cells: bottom-left origin, relative
+        to the page boundary and rotated by /Rotate.
+        """
         return [
             _to_bounding_box(tuple(bbox))
             for bbox in self._require_page_decoder().get_connected_shape_bounding_boxes(
--- a/src/parse/pdf_decoders/page.h
+++ b/src/parse/pdf_decoders/page.h
@@ -147,6 +147,10 @@ namespace pdflib
 
     void rotate_contents();
 
+    // Map a bbox from the raw content-stream space of the render instructions
+    // onto the frame of the sanitized page items (see page_frame_* below).
+    std::array<double, 4> to_page_frame(std::array<double, 4> bbox) const;
+
     bool can_reuse_sanitised_cells_for_line_cells(const decode_config& config) const;
 
     void sanitise_contents(std::string page_boundary);
@@ -211,6 +215,16 @@ namespace pdflib
     std::shared_ptr<pdf_resource<PAGE_FONTS>> acroform_fonts;
 
     pdf_render_instructions instructions;
+
+    // The render instructions stay in unrotated user space (the renderer
+    // orients its own canvas), while page_cells, page_shapes and page_images
+    // are rotated by rotate_contents() and moved to the page boundary by the
+    // dimension sanitator. Geometry queries that walk the instructions apply
+    // the same /Rotate and boundary origin, so their boxes land in the frame
+    // of the cells.
+    int page_frame_angle = 0;
+    std::pair<double, double> page_frame_delta = {0.0, 0.0};
+    std::pair<double, double> page_frame_origin = {0.0, 0.0};
 
     pdf_timings timings;
   };
@@ -557,7 +571,8 @@ namespace pdflib
             const std::array<double, 4> visible_shape_bbox =
               clipped_bbox(shape_bbox, instr.get_clip_state(), shape_visible);
 
-            if(shape_visible and bbox_intersects(bbox, visible_shape_bbox))
+            if(shape_visible and
+               bbox_intersects(bbox, to_page_frame(visible_shape_bbox)))
               {
                 return true;
               }
@@ -591,6 +606,10 @@ namespace pdflib
 
     const double tol = std::max(0.0, tolerance);
 
+    // a quarter-turn /Rotate swaps the axes between the raw instruction
+    // space and the page frame the caller asks about
+    const bool swaps_axes = (std::abs(page_frame_angle) % 180) == 90;
+
     for(const auto& instr : instructions.get_shape_instructions())
       {
         if(not shape_instruction_strokes_visible(instr)) { continue; }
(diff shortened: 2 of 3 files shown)


PeterStaar-IBM:
@wittjeff I like this approach better than #344 . We should try to have no compute/transformations done at the python level, so we know that the data-structures in C++ are consistent.

PeterStaar-IBM:
@wittjeff Another (small) thing: if we are going to have a unit test in which we verify that the text and the shape gets transformed equally, then we might as well add a bitmap too, so we cover all three cases

bharm16:
Thanks @wittjeff  — yes, I'd like to carry it in https://github.com/docling-project/docling-parse/pull/344 if that works for you.

wittjeff:
@PeterStaar-IBM thanks. Bitmap added: the test now draws a small RGB image next to the word and, for `/Rotate` 0/90/180/270, checks that the text cell, the sanitized shapes, the bitmap and the three queries all agree with one expected transform of the raw geometry. `intersects_with()` is probed with `shapes=True` and with `bitmaps=True`, at the expected boxes (must hit) and at the raw user-space location (must miss). The earlier CI red was only ruff-format on the test file, fixed in the same commit.

One unrelated thing from the same area, also noted in the description: the shape branch of the dimension sanitator sets `is_contained_in_boundary=true` in its else branch, so it never filters shapes outside the boundary. I left it alone here because it changes regression output. Happy to send a separate PR if you want it fixed.

@bharm16 works for me. The change is a single signed-off commit, 82f0dda03c7d329a5453afd45447cda9194ce422, so `git cherry-pick 82f0dda` onto #344 keeps the authorship and the DCO trailer. Drop the Python-side translation in `pdf_parser.py` when you do, since Peter wants no coordinate math at that level, and keep your test alongside. I will close this one once #344 has it. If Peter would rather merge here instead, that is fine by me too.


The CI run for the new commit is waiting for workflow approval again, if one of you can trigger it.

bharm16:
Thanks @wittjeff — I've cherry-picked your fix into [#344](https://github.com/docling-project/docling-parse/pull/344), removed the Python-side translation, and kept my original regression test alongside yours. I also added a small typing-only follow-up to use `pytest.approx` for the page-size assertion so it passes mypy. It should be ready for CI/re-review now.

